PR #48, opened 2013-05-30
Added execution parameter to raspistill.c to execute command after succesfull image safe

JiriHa:
Usefull with timelapse parameter, then exec command is run after every image save to move, upload or just work with completed image.
Also added option if in timelapse mode and timeout=0 it will disable the timeout and continue


popcornmix:
I'll close due to lack of response. Feel free to re-open if you want to rebase and continue with this.

PR #132, opened 2013-12-27
Circular buffer functionality

ghollingworth:
Added circular buffer functionality

This commit adds a circular buffer mode which allows the user to not write the video continuously to a file if they are interested in the last N seconds of video.  It records into memory continuously and only saves the last N seconds when a signal or keypress is received.

Fixed to make sure always gives valid video file!


GoldenDarts:
is it possible to make that raspivid doesn't stop when triggering -USR1 signal, only taking t seconds of video and write it into file video%04d.h264 or directly to stdout. It would be great to trigger it more than once without spending time on initializing camera every time.

JamesH65:
Certainly possible, but unlikely we would be do it. Might be worth asking on the forum if someone can make the mods for you.

We will always take a look at PR's adding new functionality.

PR #136, opened 2014-01-06
Wrong sign in if condition

ethanol100:
if (-this_delay_ms < -state->timelapse/2), -this_delay_ms is always positive,
and -state->timelapse/2 is negative, so this if is never true.  The if
condition should be if (-this_delay_ms < state->timelapse/2).


popcornmix:
@6by9 ?

6by9:
I've certainly never seen the "late" message, and the explanation of the logic error sounds sensible. I've not tested it, but I'm happy enough by inspection.

PR #166, opened 2014-04-21
Relocate vc_vchi_dispmanx_common.h include.

gamaral:
Currently vc_vchi_dispmanx.h includes vc_vchi_dispmanx_common.h which is
not distributed on install. This limits access to the ELEMENT_CHANGE_*
defines.

It's not needed in the header itself, it works just fine in the
implementation file.


Ruffio:
@gamaral can this PR be closed? It looks like it has been resolved.

Ruffio:
@popcornmix can this be merged/closed ?

JamesH65:
@popcornmix Are we wanting to merge this? Otherwise close I think - its ancient.

popcornmix:
Obviously not affecting many users, but we do provide ``interface/vmcs_host/vc_vchi_dispmanx.h`` and not ``interface/peer/vc_vchi_dispmanx_common.h`` which it tries to include, so the PR still makes sense.

PR #171, opened 2014-05-11
Add exposure=off option

aizvorski:
This should be a fixed shutter speed mode, but the behavior depends on how MMAL interprets this setting.  In any case it works, I am not sure what exactly it ends up doing but it is analogous to having awb=off option.


6by9:
@popcornmix and @JamesH65 This can be accepted now, though may need rebasing.

The firmware bug that wasn't setting digital gain to a sensible default has been fixed. (https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=96821, and "firmware: AGC tuner: Correction to setting default digital gain to x1.0" from https://github.com/raspberrypi/firmware/commit/856e2e1907a7f7058289c30268515c8cbf3fa5e3)

PR #184, opened 2014-06-29
NetBSD fixes. Includes fixes for gcc 4.8

skrll:
github newbie so I'm unsure if I should be requesting pull requests for subsets of commits.


brd:
Any thoughts on getting this in? It has been open awhile without any comment..

popcornmix:
The PR is a bit of a monster. It would be better if it were split into some smaller PRs.
Perhaps separate:
* cosmetics
* warning removal
* error handling
* functional changes, like the stack size increase, which should include an explanation for why it is necessary
* NetBSD specific makefile changes

Then it may be easier to gradually merge them in.

brd:
Sounds fair. Thank you for you reply.

I am looking at doing similar work for FreeBSD, so I will try and break it down.

Ruffio:
@brd will you break it down?

CMakeLists.txt
```diff
@@ -83,7 +83,10 @@ add_subdirectory(middleware/openmaxil)
 
 # FIXME: this directory needs a more sensible name and better
 # platform factoring
-add_subdirectory(interface/usbdk)
+
+if(NOT CMAKE_SYSTEM_NAME MATCHES "NetBSD")
```
luked99: I think this directory has been removed now entirely.

Ruffio:
Is this PR still relevant?

skrll:
If anything is missing I'll do it differently (i.e. better)

PR #211, opened 2015-01-08
Added a new camera application, raspividyuv

JamesH65:
New demo app to produce raw YUV420 video files.


JamesH65:
Well, I haven't tried extensive testing, but did get a seemingly valid
result for a 10s 720p30 clip. I suspect Linux was caching the file output
somewhat, as that rate is higher than the SD card capability...

I was going to put in my own buffering scheme (record to RAM, save behind),
but I think it's OK to rely on the Linux buffering for the moment.

PR #215, opened 2015-02-06
Added annotate and intrarefresh settings

JamesH65:
Added intra refresh type to raspivid(s)
Added an annotate parameters to allows insertion of various status' to stills or videos.


JamesH65:
I've updated the PR with the corrections suggested by 6x9.  Any other comments before merging?

(and sorry it's really two change in one - not enough time to unpick and move them around)

JamesH65:
Any objections? Will be merging by 7pm unless otherwise informed of impending disaster.

popcornmix:
No objection. I compile tested it...

JamesH65:
Good enough for me! We've been using it here, and seems OK.

PR #225, opened 2015-03-10
Raspistill/vid: More annotate options, and stereoscopic support.

6by9:
Support for stereoscopic - options 
`-3d [sbs|tb]`    selects side by side, or top/bottom  
`-3dswap`    to swap the two channels.
`-dec`   to reduce the width/height of each image by a factor of 2.

Support for more annotation options:
`-ae <text size> <text colour> <background colour>`. 
text_size is 6 to 160.
colours are YUV as hex 0xVVUUYY. Text colour only uses Y at the moment.

(Thought I'd done this pull request a while back, but obviously not).


popcornmix:
Fine by me. @JamesH65 any comments?

PR #239, opened 2015-04-22
Fix intra-refresh port parameter setting.

### LINKED ISSUE #238: mmal_port_parameter_get()/set() not idempotent for IntraRefresh.
Hi,

I was investigating https://github.com/thaytan/gst-rpicamsrc/issues/22 in gst-rpicamsrc (which is Raspivid turned into a GStreamer source element), in which using intra-refresh modes was causing an error.

Eventually, I figured out that mmal_port_parameter_set() with the MMAL_PARAMETER_VIDEO_INTRA_REFRESH_T causes problems, even though the struct is intialised with mmal_port_parameter_get() as in raspivid.

It looks like the air_ref, cir_mbs and pir_mbs fields are not populated by the call to mmal_port_parameter_get(), so they end up with whatever is on the stack. I guess in raspivid that's happening to always be 0, or some other valid value. memset'ing the struct to 0 before mmal_port_parameter_get() makes it always work for me, but it seems like the fields should be initted by the call.

thaytan:
check mmal_port_parameter_get() succeeds when getting
the default intra refresh values. It fails on older firmware.

Fixes #238

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -1604,6 +1604,12 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
 
       // Get first so we don't overwrite anything unexpectedly
       status = mmal_port_parameter_get(encoder_output, &param.hdr);
+      if (status != MMAL_SUCCESS)
+      {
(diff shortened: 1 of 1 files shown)


thaytan:
Update - fixed the indent of the comment. Not sure how that tab snuck in.

PR #240, opened 2015-04-28
Zero copy, and buildme change for Pi2

6by9:
As per commit text.
- buildme dropped into the cross compile path when building on Pi2, so didn't install.
- enable zero copy within MMAL.

### DIFF
--- a/buildme
+++ b/buildme
@@ -1,11 +1,15 @@
 #!/bin/bash
 
-if [ "armv6l" = `arch` ]; then
+if [ "armv6l" = `arch` ] || [ "armv7l" = `arch` ]; then
 	# Native compile on the Raspberry Pi
 	mkdir -p build/raspberry/release
 	pushd build/raspberry/release
 	cmake -DCMAKE_BUILD_TYPE=Release ../../..
-	make
+	if [ "armv6l" = `arch` ]; then
+		make
+	else
+		make -j4
+	fi
 	if [ "$1" != "" ]; then
 	 sudo make install DESTDIR=$1
 	else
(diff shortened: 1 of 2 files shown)


6by9:
Ping @jamesh too.

lurch:
I suspect you meant to ping @JamesH65 ? ;)

popcornmix:
Does ENABLE_MMAL_VCSM default to:
```mmal_port_parameter_set_boolean(m_vout_input, MMAL_PARAMETER_ZERO_COPY,  MMAL_TRUE)```

will that cause any issues? 
What happens when "/dev/vcsm" is not accessible by non-root user? Does it fall back to non zero copy, or fail?

6by9:
Default for MMAL_PARAMETER_ZERO_COPY is MMAL_FALSE, so this is only adding the capability for zero copy, not enabling by default.

Zero copy for MMAL_ENCODING_OPAQUE is done via a MMAL call for some reason, hence I suspect how you managed to get it working.
Zero copy for any other encoding will use /dev/vcsm, and will fail at the point the pool is created if that device is not accessible (hence our recent email thread over permissions on /dev/vcsm).

Oops over incorrect ping - thanks for amending.

popcornmix:
Okay, so this PR shouldn't break anything (even if vcsm is not accessible)?
It will only potentially break when/if MMAL_PARAMETER_ZERO_COPY is requested in future?

JamesH65:
Is vcsm now regarded as 'ok'? rather than experimental? (Not that I have any issues with this since by default it is off)

6by9:
@popcornmix Correct - no changes by default with this PR. (Stuff I'm working on will benefit from zero copy, so I was enabling it from the start).

@JamesH65 vcsm seems to work on Pi fairly happily. Whether we consider it mainstream, or just an extra feature that may be of use is open to debate. I guess @pelwell would be the one to veto this if necessary, as any support burden is likely to fall at his door.

PR #247, opened 2015-06-22
raspistill: Apply GPS info from gpsd as EXIF tags

jasaw:
Reads real-time GPS information from gpsd and apply them as EXIF tags. This is especially useful when running raspistill in timelapsed mode and GPS EXIF tags need to be updated in real-time.

### DIFF
--- a/host_applications/linux/apps/raspicam/CMakeLists.txt
+++ b/host_applications/linux/apps/raspicam/CMakeLists.txt
@@ -26,7 +26,7 @@ add_executable(raspividyuv  ${COMMON_SOURCES} RaspiVidYUV.c)
 
 set (MMAL_LIBS mmal_core mmal_util mmal_vc_client)
 
-target_link_libraries(raspistill ${MMAL_LIBS} vcos bcm_host GLESv2 EGL m gps)
+target_link_libraries(raspistill ${MMAL_LIBS} vcos bcm_host GLESv2 EGL m)
 target_link_libraries(raspiyuv   ${MMAL_LIBS} vcos bcm_host)
 target_link_libraries(raspivid   ${MMAL_LIBS} vcos bcm_host)
 target_link_libraries(raspividyuv   ${MMAL_LIBS} vcos bcm_host)
--- a/host_applications/linux/apps/raspicam/CMakeLists.txt
+++ b/host_applications/linux/apps/raspicam/CMakeLists.txt
@@ -26,7 +26,7 @@ add_executable(raspividyuv  ${COMMON_SOURCES} RaspiVidYUV.c)
 
 set (MMAL_LIBS mmal_core mmal_util mmal_vc_client)
 
-target_link_libraries(raspistill ${MMAL_LIBS} vcos bcm_host GLESv2 EGL m)
+target_link_libraries(raspistill ${MMAL_LIBS} vcos bcm_host GLESv2 EGL m gps)
 target_link_libraries(raspiyuv   ${MMAL_LIBS} vcos bcm_host)
 target_link_libraries(raspivid   ${MMAL_LIBS} vcos bcm_host)
 target_link_libraries(raspividyuv   ${MMAL_LIBS} vcos bcm_host)
--- a/host_applications/linux/apps/raspicam/CMakeLists.txt
+++ b/host_applications/linux/apps/raspicam/CMakeLists.txt
@@ -19,14 +19,14 @@ set (COMMON_SOURCES
    RaspiCLI.c
    RaspiPreview.c)
 
-add_executable(raspistill ${COMMON_SOURCES} RaspiStill.c  RaspiTex.c RaspiTexUtil.c tga.c ${GL_SCENE_SOURCES})
+add_executable(raspistill ${COMMON_SOURCES} RaspiStill.c  RaspiTex.c RaspiTexUtil.c tga.c ${GL_SCENE_SOURCES} libgps.c)
 add_executable(raspiyuv   ${COMMON_SOURCES} RaspiStillYUV.c)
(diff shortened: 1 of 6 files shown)


popcornmix:
Hmmm. The extra library dependant library is problematic.
I can no longer build firmware updates (where userland repo is cross-compiled),
and everyone's builds will be broken until they install libgps-dev.

ping @JamesH65 @6by9 for comments.

JamesH65:
That was my thought as well. GPS library isn't installed by default I
presume, so this breaks everyone's build. If there a way round this by
dynamical testing if the library is present?

6by9:
Needs to be build time checking of the library's presence, otherwise it won't find the headers.
My CMake foo isn't up to it, but that is something that the old ./configure scripts were always good at, so I would have expected CMake to have something similar. Google seems to have some likely looking candidates though.

JamesH65:
Still means the individual has to rebuild the software if they want GPS.
Can we not use dynamic loading to test if the library is present, then use
it if so? One build to rule  them all and in the darkness bind them. Sort
of.

popcornmix:
I assume you need an attached gps receiver to make use of this? Seems a pretty niche use case for what is primarily a demo app.

6by9:
If you don't amend the CMake config then it will still break builds until you install libgps-dev, as the '#include "gps.h" will fail. Dom is right that currently there are no additional dependencies for building userland. 
You could do both CMake mods and dynamic loading, so basic builds still work and more advanced builds will try GPS if it exists, but not fail if it doesn't. Breaking either side would be bad.

Yes, it's a demo app, but receivers appear to be <£20, so others may be interested if the functionality were hooked in. I'd suggested he create a PR on https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=102617 as otherwise the code was just going to get lost on the forum, but happy if you want to push back on it.

JamesH65:
Can get away without gps.h...


    int main(int argc, char **argv) {
        void *handle;
        double (*cosine)(double);
        char *error;

        handle = dlopen ("/lib/libm.so.6", RTLD_LAZY);
        if (!handle) {
            fputs (dlerror(), stderr);
            exit(1);
        }

        cosine = dlsym(handle, "cos");
        if ((error = dlerror()) != NULL)  {
            fputs(error, stderr);
            exit(1);
        }

        printf ("%f\n", (*cosine)(2.0));
        dlclose(handle);
    }

popcornmix:
If it can be done through dlopen/dlsym without a dependency during build, and handles absence of gps lib gracefully then I won't object.

jasaw:
Using dlopen and dlsym is definitely not a problem, but that means duplicating information in gps.h. In the unlikely event of gpsd API changes, raspistill will break. If everyone's ok with that, I'll make the changes.

JamesH65:
Thanks, I think removing the build time reliance on gps library is the best
approach.

jasaw:
I've updated the code to only load libgps.so.20 at runtime if gps option is used.
gps.h from libgps is duplicated in raspicam folder and can be easily replaced in the unlikely event of libgps API changes.

jasaw:
Currently, if it losses GPS lock, it does *not* write GPS EXIF, i.e. it does *not* cache GPS data. There's more discussion here https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=102617
Is everyone happy with the way it is now?

One improvement for dealing with losing GPS lock is to estimate its current latlong based on last known latlong, speed, and track, but estimation is only good for 5 seconds.

JamesH65:
I've not tested this  -has anyone else? If it works as advertised it could be merged.

jasaw:
It currently does not work with Raspbian Jessie. Jessie ships with libgps21, but this patch is expecting libgps20. There are changes in the data structures going from libgps20 to libgps21.

jasaw:
Looks like it's *not* going to be easy to support multiple versions of libgps.
One way is to somehow wrap each version of gps.h to avoid struct / macro name clash, but this is a nightmare to maintain, i.e. need to update raspistill each time libgps API changes.

What do you guys prefer?

6by9:
(At work so can't look in too much detail at the moment)
So it looks like we're jumping form gpsd-3.6 in libgps20, to gpsd-3.15 in libgps21. API version 5.0 to 6.1, so yes there are incompatible API changes :-(

I can't see a way to find out GPSD_API_MAJOR_VERSION for the installed version - that seems a slight omission in the library. I guess you're explicitly loading libgps.so.20 so can derive it from there, but my brain is failing to see how a program compiled with -lgps can run on a different system that may have a different libgps installed. They can't really expect everyone to recompile the app, can they?

3.6 was May 2012, 3.15 was June 2015. If they update once every 3 years, then it'll be annoying when the packages get updated, but not a huge overhead.

I know that's not an answer, just a brain dump of my thoughts.
I'd be tempted to ditch libgps20 and only support libgps21. It's not a feature that has been available before to rapistill, so requiring a Jessie release seems reasonable (comment it in the docs, and poss a friendly error if GPS requested and wrong library found).

6by9:
@JamesH65 @popcornmix Any thoughts?

popcornmix:
Agreed that only supporting Jessie for a new feature is fine.
It should fail gracefully on Wheezy (i.e. you can't add gps info, but otherwise it works).

JamesH65:
Agreed. It's a nice feature if it can be incorporated transparently.

jasaw:
Question: EXIF tags are added by calling mmal_port_parameter_set, but when I lose GPS lock, how do I unset the EXIF tags so the tags don't stick around in the subsequent photos?

6by9:
Set the same way, but leave it blank after the =, eg
`GPS.GPSLatitudeRef=`

jasaw:
Superseded by pull-request #286

PR #248, opened 2015-06-24
Added support for network and fflush to decrease latency.

Kixunil:
These changes allow users to stream video directly over network with lower latency (bypassing pipes and buffers). If file name has format tcp://host:port then TCP connection to host:port is opened. Similarly if udp:// is used, UDP packets with data are sent to host:port.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -56,6 +56,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <memory.h>
 #include <sysexits.h>
 
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
 #define VERSION_STRING "v1.3.12"
 
 #include "bcm_host.h"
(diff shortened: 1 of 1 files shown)


JamesH65:
Would it be possible to use the same bracketing style as the rest of the code please (Allman).

popcornmix:
Regularly calling fflush will reduce performance for the usual case (writing to sdcard) where we don't care about latency.

JamesH65:
I'd suggest it would be worth putting the fflush as an optional parameter
just for those who need it..

Kixunil:
Sorry to come back so late, I forgot about this PR. I fixed the braces and made optional switch to enable flushing. In case of other issues feel free to write me.

Kixunil:
Sorry, I used tabs in vim. Fixed it.

popcornmix:
Can you squash all commit to a single one and force push.
@JamesH65 @6by9 okay by you?

Kixunil:
@popcornmix Done, is it OK?

Kixunil:
Yeah, I prefer tabs and I'm very accustomed to using them. Sorry.

JamesH65:
Using "f" as the short command line parameter conflicts with the preview -f
option - I'd suggest changing to -fl which I think is free. You need to
cross check with RaspiPreview.c, Raspistill/vid.c and RaspiCamControl.c to
find a 'spare' option.

6by9:
Content looks fine to me other than James having picked up on the clash of -f in the option.

Kixunil:
@JamesH65 I didn't know that options should be conflict-free across different tools. I just changed it to -fl, as you suggested.

JamesH65:
They are not different tools, as things like RaspiPreview are used across
multiple tools. Hence the requirement to keep things unique.

Kixunil:
@JamesH65 OK, thanks for explanation!

PR #258, opened 2015-09-20
Proposing VC_DISPLAY env variable

maxnet:
It would be nice if there was a standardized way to instruct applications to use a certain display, without having to patch the application's source code.
Proposing a VC_DISPLAY environment variable for this purpose.

To force display on HDMI one could then execute:

```
VC_DISPLAY=5 any-egl-application
```

Similar to the standardized DISPLAY variable used by X11 applications.


popcornmix:
Thanks - that sounds useful.

popcornmix:
Latest rpi-update firmware contains this.

AaronBottegal:
This was very useful for me and very easy to use. In my case, I did DSI over HDMI on a Pi 3. Worked like a charm.

PR #259, opened 2015-09-23
Update video.c

fredericofs:
Remove the timout, to avoid the application quits before the ending of some videos.
(Application used to hang before last commit, now it quits, either, the only way to watch some video files till the end is to remove timeout completely like I am suggesting now, or to incrase it, but I dont now what should be better pratice.


popcornmix:
This is now included. Thanks.

PR #264, opened 2015-10-28
allow annotation string to contain strftime parameters like %Y-%m-%d %X

fullergalway:
This small change allows user to include strftime parameters in the annotation.

    raspivid -t 5000 --annotate "Camera 1 %Y-%m-%d %X" -o kewl.mp4

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.c
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.c
@@ -692,7 +692,7 @@ int raspicamcontrol_parse_cmdline(RASPICAM_CAMERA_PARAMETERS *params, const char
       }
       else
       {
-         params->enable_annotate = ANNOTATE_USER_TEXT;
+         params->enable_annotate |= ANNOTATE_USER_TEXT;
          //copy string char by char and replace "\n" with newline character
          unsigned char c;
          char const *s = arg2;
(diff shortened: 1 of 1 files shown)


JamesH65:
Looks OK to me, but can you update the online documentation as well please - that also in github here https://github.com/raspberrypi/documentation/blob/master/raspbian/applications/camera.md

JamesH65:
Dom, 6x9, any objections to merging this?

popcornmix:
Okay by me. @6by9 happy?

6by9:
Ish. 
% hasn't had any meaning in this string until now, so would have been displayed as is.
After this change you'll get different behaviour if your string happened to include a % that strftime knows (you'd need %% to guarantee a % sign rather than a formatted value).

Personal preference would be for this to be under a new ANNOTATE_xxx bit flag within settings so users had to consciously select the behaviour, but happy to be downvoted as it is going to be a rare case.
(And I am aware that the overloading of "-a" is horrid and catches out the unwary - setting the string second will override the value set by a "-a bitmask" parameter before it. Don't need git blame for that one though, do we ;-) )

JamesH65:
Good spot on %. Not sure how to deal with that neatly and keep the same
system. I sort of like the idea of a new ANNOTATE bit flag (notwithstanding
it's horrid, not sure what possessed me). TBH, I suspect the extra %
requirement to end users wouldn't be a big issue.

6by9:
Actually there's an easy fix for ```-a <bitmask> -a "String"```. https://github.com/fullergalway/userland/blob/5b48fd85eb799ab66af4a269ccf8e4cdf2d05e9f/host_applications/linux/apps/raspicam/RaspiCamControl.c#L695 becomes
 params->enable_annotate |= ANNOTATE_USER_TEXT;
You might get USER_TEXT set when you hadn't specified it, but why bother with the string if you don't want it displayed!

Reduce the chance of collateral damage by requiring ANNOTATE_TIME_TEXT or ANNOTATE_DATE_TEXT to be set and find a % in the string? (It does then need to then block those if clauses from adding the time/date with the default formatting). 
People complain over any regression, even if there is a simple workaround - "it worked last week!" See the saga over correcting the display orientation.

fullergalway:
Right I made that change but sorry amn't near a compiler right now. I'll update the documentation merge request accordingly.

6by9:
Thank you. Looks OK to me (assuming it does compile!), and should reduce the risk of complaints to a level that I don't care.
I suspect popcornmix may request that the two patches get squashed in to one though.

popcornmix:
Yes, please squash commits into one, and force push to update the PR.

fullergalway:
Squashed commits.

PR #266, opened 2015-11-04
determine better whether annotation is a bitmask or string

fullergalway:
Before this change the parameter "--annotate 12ab" annotates with a date.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.c
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.c
@@ -692,7 +692,7 @@ int raspicamcontrol_parse_cmdline(RASPICAM_CAMERA_PARAMETERS *params, const char
       }
       else
       {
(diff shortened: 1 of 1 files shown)


pelwell:
Assuming that there isn't a better way to indicate whether an annotation is a string or a date other than by examining the supplied value, I think there is a simpler implementation.

Since sscanf returns the number of values assigned, couldn't you just replace:
```
if (isdigit(*arg2))
{
```
with:
```
char dummy;
if (sscanf(arg2, "%u%c", &params->enable_annotate, &dummy) == 1)
```
?

fullergalway:
Or maybe, to avoid polluting params->enable_annotate in the case of 12ab..

    char dummy;
    unsigned int bitmask;
    if (sscanf(arg2, "%u%c", &bitmask, &dummy) == 1)

?

pelwell:
Yes, that looks better.

PR #267, opened 2015-11-04
determine better whether annotation is a bitmask or string

fullergalway:
Before this change the parameter "--annotate 12ab" annotates with a date.

Replaces pull request #266 incorporating suggested changes.


pelwell:
Looks OK to me.

JamesH65:
OK from me. Wish I hadn't written the code like that originally, its a bit
of a horrible way of doing it!

Issue #274, opened 2015-12-05
rastpstill with timelapse and datetime does not include Year


mfn:
The `--help` states:
`-dt, --datetime : Replace frame number in file name with DateTime (YearMonthDayHourMinSec)`

Using this command line, however, the generated files do not include the year:

`raspistill --width 1920 --height 1080 --encoding png --timelapse 10000 --datetime --timeout 0 --output image_%d.png`

Generated files:
```
pi@raspberrypi:~/tmp $ ls -l
total 680
-rw-r--r-- 1 pi pi 39075 Dec  5 23:15 image_1205231531.png
```
It should be `image_20151205231531.png`

My c-fu is limited, but the code in [RaspiStill.c](https://github.com/raspberrypi/userland/blob/daad93e972f2332555ca2dc7b5a27a89d52572d5/host_applications/linux/apps/raspicam/RaspiStill.c#L1803-L1819) looks like it simply doesn't take the year into account:
```C
                if (state.datetime)
                {
                   time_t rawtime;
                   struct tm *timeinfo;

                   time(&rawtime);
                   timeinfo = localtime(&rawtime);

                   frame = timeinfo->tm_mon+1;
                   frame *= 100;
                   frame += timeinfo->tm_mday;
                   frame *= 100;
                   frame += timeinfo->tm_hour;
                   frame *= 100;
                   frame += timeinfo->tm_min;
                   frame *= 100;
                   frame += timeinfo->tm_sec;
                }
```

6by9:
A signed 32bit int will take a max value of 2147483647, which only gives enough range for MMDDHHMMSS. I suspect the original author found that things weren't quite as they expected and just trimmed off the year.

If you want to create a patch for this, then you could convert frame to int64_t (may need to include stdint.h). However that will also mean the format string for the filename has to change, and that is provided by the user so will break every single app that uses the datetime option. 
Backwards compatibility must be maintained here as the app is widely used, therefore the only real fix is to update the help string - sorry.

mfn:
> Backwards compatibility must be maintained here as the app is widely used, therefore the only real fix is to update the help string - sorry.

No problem, I completely understand this!

I guess producing a working patch is out of my league (feature wise); however if someone would do that, maybe an approach would be to add a new option, like `--datetime-for-real` (just kidding, but the idea ...)

pelwell:
You could always rename it afterwards. It would be trivial to write a script that would do the job, and as long as you run it within a year of capture you'll get the right result.

[...]

Issue #244, opened 2015-05-31
raspivid --help output is printed to stderr instead of stdout


derekenos:
This prevents pagers like `less` and `more` from working as expected when the height of the content exceeds that of the terminal window.

Example:

`raspivid --help | less` displays only the last page of content:
![screenshot - 05312015 - 02 38 14 pm](https://cloud.githubusercontent.com/assets/585182/7903203/f1628e82-07a2-11e5-92a6-81368852eb69.png)

Redirecting `stderr` to `stdout` allows `less` to work as expected, `raspivid --help 2>&1 | less`:
![screenshot - 05312015 - 02 43 45 pm](https://cloud.githubusercontent.com/assets/585182/7903234/7628e38c-07a3-11e5-992b-f2fe6659554b.png)

6by9:
Feel free to sort out a patch and make a pull request.
You are right that all the fprintf lines are set to stderr, when stdout would probably be more useful.

Likewise I believe all the vcos_log_xxx lines are going to stderr. A quick tweak to https://github.com/raspberrypi/userland/blob/master/interface/vcos/pthreads/vcos_pthreads.c#L334 is likely to fix that, although that affects all users of vcos logging on the ARM, which includes the MMAL IPC framework and the likes.

popcornmix:
Be aware that raspivid supports output of encoded data to stdout and so you need to be careful not to mix logging output with that.

JamesH65:
Following on from popcornmix, all output was pushed to stderr in order to
enable the encoded output to stdout. However, I see no reason off hand why
the help cannot go to stdout, since you won't be using them at the same
time as encoded output.

6by9:
Change merged that sends the help text to stdout instead of stderr.
Please retest and close, unless you think there is something further to do.

6by9:
@popcornmix Do you want to close this? The issue is fixed, and the original poster seems to have gone away.

derekenos:
Sorry, I've been meaning to respond. Thanks for the fix!

Issue #236, opened 2015-03-27
Remove RaspiCamDocs.odt ?


lurch:
Now that all the [documentation](http://www.raspberrypi.org/documentation/) is hosted in it's own [repo](https://github.com/raspberrypi/documentation/), `host_applications/linux/apps/raspicam/RaspiCamDocs.odt` in this userland repo is becoming more and more out of date, which is especially evident when you compare the histories of [raspberrypi/documentation/raspbian/applications/camera.md](https://github.com/raspberrypi/documentation/commits/master/raspbian/applications/camera.md) and [raspberrypi/userland/host_applications/linux/apps/raspicam/RaspiCamDocs.odt](https://github.com/raspberrypi/userland/commits/master/host_applications/linux/apps/raspicam/RaspiCamDocs.odt).

It would probably make sense to just remove `host_applications/linux/apps/raspicam/RaspiCamDocs.odt` and simply treat http://www.raspberrypi.org/documentation/raspbian/applications/camera.md as the definitive camera-apps documentation? (Or, in case any third-party sites are linking directly to RaspiCamDocs.odt, another option would be to simply delete its current contents, and change it to say something like "The current documentation for the camera applications can be found at http://www.raspberrypi.org/documentation/raspbian/applications/camera.md" ?)

JamesH65:
Agreed. Now we have the decent repo based docs the ODT is no longer
necessary. I think your recommendation to change the text to point to the
online docs is the best option.

XECDesign:
+1

popcornmix:
A PR would be good.

6by9:
Change merged which removes RaspiCamDocs.odt. 
I didn't add a reference to the official page - linking to any file in a github repo would be rather brave, so I don't see it as being an issue that the file has now gone.

6by9:
Please close unless you think there is anything further to be done.

Issue #220, opened 2015-02-16
vcos_timer_delete and other functions incorrectly declared inline


rcombs:
This produces giant piles of compiler warnings:
```
/opt/vc/include/interface/vcos/vcos_timer.h:112:6: warning: inline function ‘vcos_timer_delete’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_timer.h:109:6: warning: inline function ‘vcos_timer_reset’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_timer.h:102:6: warning: inline function ‘vcos_timer_cancel’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_timer.h:95:6: warning: inline function ‘vcos_timer_set’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_timer.h:80:15: warning: inline function ‘vcos_timer_create’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_mem.h:93:15: warning: inline function ‘vcos_get_free_mem’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_thread.h:269:6: warning: inline function ‘vcos_thread_resume’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_thread.h:262:5: warning: inline function ‘vcos_thread_running’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_thread.h:251:15: warning: inline function ‘vcos_change_preemption’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_thread.h:234:6: warning: inline function ‘vcos_thread_relinquish’ declared but never defined [enabled by default]
/opt/vc/include/interface/vcos/vcos_thread.h:186:15: warning: inline function ‘vcos_thread_get_affinity’ declared but never defined [enabled by default]
```

popcornmix:
Are you building on Pi or cross compiling?
Debug or release build?

rcombs:
Building on the Pi; debug build. The relevant functions are defined in here:
https://github.com/raspberrypi/userland/blob/4333d6d023a1f0dcc763d0706a1fe6523f9b6005/interface/vcos/pthreads/vcos_pthreads.c#L874
but declared as inline in here:
[...]

PR #277, opened 2015-12-31
Various open issue fixes

6by9:
Issues #274 #236 #244 and #220 all hopefully addressed.
(Bored, so wanted to knock a load of issues on the head).

Only the VCOS changes are potentially contentious, but seeing as most the functions weren't implemented before, it doesn't make any real difference. Please sanity check first though :-)


popcornmix:
Thanks - look good to me.

Issue #142, opened 2014-02-07
raspistill -o filename produces segfault as soon as there is a % in the filename


msperl:
Hi!

```raspistill -o filename-%s.jpg```
triggers a segfault as long as the pattern is not a valid sprintf pattern with an integer argument.

I was actually trying to run a strftime format like "file-%Y%m%d-%H:%M:%S.jpg" as the filename when I found this.

Maybe you should implement some other replacement scheme, that does not "hickup" with a segfault or at least do some validity checks on the filename patttern given...

Martin

Ruffio:
@JamesH65 whats the status on this one?

6by9:
For timelapse mode the frame number is inserted in the filename by using sprintf(filename, template, frame_num), where template is the filename passed from the command line. The expectation is that the user will insert %d in the string at the appropriate place. 
Insert %s and it tries to treat the frame number as a string, hence the segfault. 

Yes, the code could do some validity checking to ensure the supplied string either has %d or %%, rejecting any other %<char>. Yes it could expand all invalid occurrences of %<char> with %%<char>, but seeing as it is only demo code that seems a little excessive.

msperl:
Is it really just demo code?

I guess people are using the command-line inside of shell-scripts and such, so it should be stable and not segfault...

JamesH65:
Indeed. I wrote it, put in rudimentary error checking, but it's never been
claimed it's production ready - hence demo code, and hence I put in quite a
few comments in the code to make it easier to understand. I was expecting
by now that people would have taken the code and either written their own
versions or hardened the base code. No such luck. If I have time I might be
able to spend some effort on it, but I would prefer people just make
corrections/additions, send in pull requests, where they will be peer
reviewed and accepted as and when.

6by9:
I have a patch for this on my laptop that should sanity check the filename. 
Not checked it yet as I didn't have a Pi to hand, and can't cross-compile it under Cygwin (or not that I've been bothered to solve).

My update was just that I was bored so going through issues trying to knock a bundle on the head (there are a load where the reporter hasn't responded to questions, and/or have been fixed but not closed).

6by9:
Patch pushed to my branch and PR created as https://github.com/raspberrypi/userland/pull/279

6by9:
PR merged. Please close if happy.

Issue #34, opened 2013-05-21
raspicam --ev -10..10 is not EV


towolf:
I don’t know what this is doing but what it is setting is not EV, i.e., log_2 scale. It’s quite small increments. Please clarify

Ruffio:
@towolf is this still an issue?

6by9:
PR #279 now merged which adds help text stating that EV is in steps of 1/6 stop.
Please close unless you think there is something further to do.

Ruffio:
@towolf is this still an issue? It looks like it has been solved.

towolf:
No

PR #279, opened 2016-01-05
Fixes for #34 and #142

6by9:


6by9:
@JamesH65 and @msperl please review and test the filename sanity checking change.

6by9:
@msperl - patch updated to accept %<numerics>d, as well as %%, or %d.
Tested with and allows:
file%%%%
file%%%%.jpg
file%%%08d
file%%%%08d (valid, but doesn't use the frame number argument, just gives "file%%08d")

Rejects:
file%%%08x
file%s

(Probably shouldn't have forced updated this branch, but saves a squash later)

msperl:
I guess "%.8d" is also valid in principle, but I guess nobody would use that - and if someone does then the workaround is to use "%08d".

At least it does not segfault any longer...

popcornmix:
Thanks.

PR #280, opened 2016-01-06
GCC5 fixes

TheSin-:
This is just a small patch That I have been using to build my userland deb files with gcc5, most of it is pretty safe and harmless, mostly printf (and friends) to fputs and such.

The inline declarations is the only one I'm not sure would be backwards compat and none of my build systems have gcc4.9 or are none GNUC to test on.

Be nice to get this accepted so I no longer have to patch my builds anymore.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStill.c
+++ b/host_applications/linux/apps/raspicam/RaspiStill.c
@@ -350,7 +350,7 @@ static void dump_status(RASPISTILL_STATE *state)
    for (i=0;i<next_frame_description_size;i++)
    {
       if (state->frameNextMethod == next_frame_description[i].nextFrameMethod)
-         fprintf(stderr, next_frame_description[i].description);
+         fprintf(stderr, "%s", next_frame_description[i].description);
    }
    fprintf(stderr, "\n\n");
 
--- a/host_applications/linux/apps/raspicam/RaspiStillYUV.c
+++ b/host_applications/linux/apps/raspicam/RaspiStillYUV.c
@@ -267,7 +267,7 @@ static void dump_status(RASPISTILLYUV_STATE *state)
    for (i=0;i<next_frame_description_size;i++)
    {
       if (state->frameNextMethod == next_frame_description[i].nextFrameMethod)
-         fprintf(stderr, next_frame_description[i].description);
+         fprintf(stderr, "%s", next_frame_description[i].description);
    }
    fprintf(stderr, "\n\n");
 
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -400,7 +400,7 @@ static void dump_status(RASPIVID_STATE *state)
    for (i=0;i<wait_method_description_size;i++)
    {
       if (state->waitMethod == wait_method_description[i].nextWaitMethod)
-         fprintf(stderr, wait_method_description[i].description);
+         fprintf(stderr, "%s", wait_method_description[i].description);
    }
    fprintf(stderr, "\nInitial state '%s'\n", raspicli_unmap_xref(state->bCapturing, initial_map, initial_map_size));
    fprintf(stderr, "\n\n");
--- a/host_applications/linux/apps/raspicam/RaspiVidYUV.c
+++ b/host_applications/linux/apps/raspicam/RaspiVidYUV.c
@@ -300,7 +300,7 @@ static void dump_status(RASPIVIDYUV_STATE *state)
    for (i=0;i<wait_method_description_size;i++)
    {
       if (state->waitMethod == wait_method_description[i].nextWaitMethod)
-         fprintf(stderr, wait_method_description[i].description);
+         fprintf(stderr, "%s", wait_method_description[i].description);
    }
    fprintf(stderr, "\nInitial state '%s'\n", raspicli_unmap_xref(state->bCapturing, initial_map, initial_map_size));
    fprintf(stderr, "\n\n");
--- a/host_applications/linux/apps/raspicam/gl_scenes/models.c
+++ b/host_applications/linux/apps/raspicam/gl_scenes/models.c
@@ -255,7 +255,7 @@ static int load_wavefront_obj(const char *modelname, WAVEFRONT_MODEL_T *model, s
                m->material_index[m->num_materials] = (pf-qf)/3;
                m->num_materials++;
             }
-         } else { printf(s); vc_assert(0); }
+         } else { printf("%s", s); vc_assert(0); }
          break;
       case 'g': vc_assert(strncmp(s, "g ", sizeof "g "-1)==0); break;
       case 's': vc_assert(strncmp(s, "s ", sizeof "s "-1)==0); break;
@@ -305,7 +305,7 @@ static int load_wavefront_obj(const char *modelname, WAVEFRONT_MODEL_T *model, s
                *pf++ = pp[3*i+0]; *pf++ = pp[3*i+1]; *pf++ = pp[3*i+2];
                *pf++ = pp[3*(i+1)+0]; *pf++ = pp[3*(i+1)+1]; *pf++ = pp[3*(i+1)+2];
             }
-         } else { printf(s); vc_assert(0); }
+         } else { printf("%s", s); vc_assert(0); }
          break;
       default: 
          printf("%02x %02x %s", s[0], s[1], s); vc_assert(0); break;
--- a/host_applications/linux/apps/smem/smem.c
+++ b/host_applications/linux/apps/smem/smem.c
@@ -192,7 +192,7 @@ int main( int argc, char **argv )
    int  opt;
    int  opt_alloc = 0;
    int  opt_status = 0;
-   uint32_t alloc_size;
+   uint32_t alloc_size = 0;
    int  opt_pid = -1;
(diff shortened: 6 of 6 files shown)


JamesH65:
Why did you need to change the fprintf to fputs?

TheSin-:
gcc5 will error with format-security if you use printf without an params.  We likely could just suppress it to a warning but why not just fix it.

TheSin-:
I'm not a C expert, I'm just posted what I had to do to get things to compile based not eh errors I was getting, I have been using these patches for some time so maybe I test without the inline patch, like I said int eh PR the inline one is the only one I'm not 100% sure of.  [f]printf -> [f]puts I'm sure will be 100% fine backwards and I know it's how it should be.  The uninitialized is just setting a default value that won't hurt anything and silences the compiler.  The inline I know I did need it but I will not claim I understand it.  I just know the results are working from all my testing though I do not have a raspicam to test with.

I'm using all these patches on my build scripts from https://github.com/TheSin-/rpi-img-builder which uses my own repo where I have proper deb files for the firmware/kernel and userland packages.  Since this repo is planned to be a production repo in 2016 at some point I'm just trying to clean up patches where I can.

@6by9 I'll do a test build right now without that patch and post the error is any.

TheSin-:
@6by9 the inline patch is no longer required to build, so I'm removing it, thanks.

JamesH65:
I'm still not understanding why the fprintf change is needed - as I
understand it, fprintf(stderr, "some string"); is perfectly valid, and
googling for similar dug up nothing. Note, I'm not saying the patch is
wrong, I just don't understand why this particular change is needed.

JamesH65:
Typical - just found this

-Wformat-securityIf -Wformat is specified, also warn about uses of format
functions that represent possible security problems. At present, this warns
about calls to printf and scanf functions where the format string is not a
string literal and there are no format arguments, as in printf (foo);. This
may be a security hole if the format string came from untrusted input and
contains ‘%n’. (This is currently a subset of what -Wformat-nonliteral warns
about, but in future warnings may be added to -Wformat-security that are
not included in -Wformat-nonliteral.)

However, the format string in our case IS a string literal, so this warning
shouldn't be output anyway., AIUI.

pelwell:
It's not, it's just inefficient, but `printf(s);` is a disaster waiting to happen.

TheSin-:
it's a debian/ubuntu thing mostly, it's a hardening flag.  see

https://wiki.ubuntu.com/ToolChain/CompilerFlags#A-Wformat_-Wformat-security

but if you do a web search you'll see other are having the same issue with userland and gcc5+

JamesH65:
Hmm, looks like a GCC bug since I think all the occurrences actually use
string literals (But I don't have GCC5 to check), which should avoid the
warning.

popcornmix:
It doesn't sound right that the canonical
```printf("Hello world\n");```
is now not acceptable.

**Comment by pelwell** [comment]
See https://github.com/TheSin-/userland/commit/6587ef83e1bc6617119937d500f87a6623b71ff9#diff-0d6f6c0efeae3339eee94eb32cb28648R258 - I didn't make up that example.

**Comment by JamesH65** [comment]
Think that was one of the Tim's !! That's certainly an issue.

**Comment by Ferroin** [comment]
@popcornmix I don't know where you might have learned C, but every example I've ever seen uses print("Hello world\n"), not printf().

**Comment by pelwell** [comment]
Since when has print() been part of the standard C library?

**Comment by Ferroin** [comment]
Also, we should consider ourselves lucky that this is all that broke.  I've seen a number of things that had much more complicated issues than this to deal with, including seemingly random link errors that only happened with GCC 5 (and weren't related to the change to gnu11 as the new default standard).

**Comment by TheSin-** [comment]
@pelwell a simple printf(s) might be fine, I just found all f?printf and if they didn't have params I changed them, I started one at a time but I was getting tired of the long process of waiting for failure so I just replaced them all after like 4 attempts of one at a time ;)  Figured using puts wouldn't hurt even if it was over cautious.

**Comment by Ferroin** [comment]
@pelwell.  Maybe not part of the official standard, but there are a lot of systems that do include a simple print function (which is usually a wrapper for print f internally, but you get no complaints about that for some reason from GCC on those systems).

**Comment by Ferroin** [comment]
And in retrospect, I realize that I originally learned C for embedded development (and then for systems work on Plan 9), so I feel a bit like an idiot now...

**Comment by JamesH65** [comment]
I think the only place that actually needs the change is the one in
model.c, (that's from a read through - may have missed one).

My concern is mixing things up - we use (f)printf everywhere else,
introducing a different function in what is demo code might be confusing to
beginners. And don't get me started on the reversal of the parameters. What
were they thinking?

**Comment by TheSin-** [comment]
@JamesH65, one sec I'll build it without the patch and post the error, I likely should have done that in the PR honestly.

**Comment by popcornmix** [comment]
I'm quite happy for
``printf(s);``
to be fixed. My preference would be for
``printf("%s", s);``
But I'll need convincing that changing
``printf("string literal\n")``
is useful.

**Comment by TheSin-** [comment]
@popcornmix I didn't think about using ```printf("%s", s);``` over ```puts(s);``` but would that add an extra/superfluous  percent expansion?  Not arguing actually asking for my own knowledge.

**Comment by TheSin-** [comment]
@JamesH65 @popcornmix here is the error without the patch

```
[ 83%] Building C object host_applications/linux/apps/raspicam/CMakeFiles/raspividyuv.dir/RaspiVidYUV.c.o
/root/rpi/userland/raspberrypi-userland/host_applications/linux/apps/raspicam/RaspiVidYUV.c: In function ‘dump_status’:
/root/rpi/userland/raspberrypi-userland/host_applications/linux/apps/raspicam/RaspiVidYUV.c:303:10: error: format not a string literal and no format arguments [-Werror=format-security]
          fprintf(stderr, wait_method_description[i].description);
          ^
cc1: some warnings being treated as errors
host_applications/linux/apps/raspicam/CMakeFiles/raspividyuv.dir/build.make:134: recipe for target 'host_applications/linux/apps/raspicam/CMakeFiles/raspividyuv.dir/RaspiVidYUV.c.o' failed
```

**Comment by popcornmix** [comment]
If you are calling printf then performance isn't key. I think sticking with a function that everyone is familiar takes precedence.
The subtle distinction of puts/fputs and newlines is something few people will know.

**Comment by TheSin-** [comment]
okay, makes sense what me to redo this PR to use that method instead? I don't mind at all as long as i'm helping.

**Comment by popcornmix** [comment]
Yes, that exmaple is bad. Replace with
``fprintf(stderr, "%s", wait_method_description[i].description);``
But drop the ones that just use a string literal.

**Comment by TheSin-** [comment]
okay I'll make that change locally and test it first, and then I'll amend this PR.

**Comment by JamesH65** [comment]
Looks like  the compiler doesn't know that those description strings are
(or are supposed to be) literals, whch is understandable. Those do need to
change to popcornmix's suggestion ("%s", s).

**Comment by TheSin-** [comment]
@popcornmix here is the printf("string") proof

```
[ 98%] Building C object host_applications/linux/apps/raspicam/CMakeFiles/raspistill.dir/gl_scenes/models.c.o
/root/rpi/userland/raspberrypi-userland/host_applications/linux/apps/raspicam/gl_scenes/models.c: In function ‘load_wavefront_obj’:
/root/rpi/userland/raspberrypi-userland/host_applications/linux/apps/raspicam/gl_scenes/models.c:258:10: error: format not a string literal and no format arguments [-Werror=format-security]
          } else { printf(s); vc_assert(0); }
          ^
/root/rpi/userland/raspberrypi-userland/host_applications/linux/apps/raspicam/gl_scenes/models.c:308:10: error: format not a string literal and no format arguments [-Werror=format-security]
          } else { printf(s); vc_assert(0); }
          ^
```

**Comment by popcornmix** [comment]
That is not a literal. If the first parameter to printf is a variable "s" then replace with:
```printf("%s", s);```
If it is a literal - i.e.words in quotes, e.g.
```printf("Hello world\n");```
then leave it as it is.

**Comment by popcornmix** [comment]
I'm happy with these changes. @TheSin- can you squash (``git rebase -i HEAD~6``) and force push?

**Comment by JamesH65** [comment]
Fine for me too.

**Comment by TheSin-** [comment]
sure, sorry was just setting it up for merging, since I saw there was a commit since I started this, I'll rebase it right now.

**Comment by 6by9** [comment]
Looks good to me (as long as we don't lose my commit merged this morning).

**Comment by TheSin-** [comment]
man I'm sorry I hate rebasing, let me try that again I really thought I had it right that time.

**Comment by TheSin-** [comment]
okay I'm just making it worst now, I think I'd jus like to nuke this PR and make a new one is that okay I know it sucks to loose all this convo, but I'm not 100% with git, specially with rebase obviously.

**Comment by popcornmix** [comment]
Assuming upstream is name of raspberrypi/userland repo and you have a copy of your changes:
```
git fetch upstream
git reset --hard upstream/master
[ add squashed commits ]
git push -f
```
should fix it up.

6by9:
Copy of your changes taken either with ```git format-patch```, or if you make a note of the commit hashes, you can ```git cherry-pick <hash>``` on to your clean branch.

TheSin-:
@6by9 thank you so much it was the cherry-pick commands I needed, I think I have it now just checking everything now

6by9:
Looks good to me.

popcornmix:
Thanks.

PR #283, opened 2016-01-22
Fix strict-aliasing

atupone:


popcornmix:
Looks correct but is a bit ugly. I think doing the memcpy on a separate line might help a bit.
@pelwell @6by9 any thoughts?

pelwell:
I think it's a shame, at least for the vchiq case, since header->data is guaranteed to be 4-byte aligned. But putting that aside, perhaps a function to read an integer from a char pointer might be a cleaner way of fixing this without having to change the indentation.

atupone:
It's ugly :( , but it is really optimized, if you the compiled code

6by9:
Nothing to add, though very ugly.

I'm assuming the vcilcs_out.c include is just for a missing prototype rather than aliasing?

atupone:
yes the include is for that

atupone:
Maybe something like this

uint32_t makeCurrentId = EGLINTMAKECURRENT_ID;
if (thread->merge_pos == CLIENT_MAKE_CURRENT_SIZE && !memcmp(thread->merge_buffer,    
                                                                                                                   &makeCurrentId,
                                                                                                                    sizeof(makeCurrentId)))
is more acceptable ?

PR #287, opened 2016-02-02
RaspiStill: Fix compiler warnings

jasaw:
Fix RaspiStill compiler warnings:
- discards const qualifier
- implicit declaration of function isdigit


6by9:
Looks fine to me.

Issue #291, opened 2016-02-15
build fails with clang 3.8.0 on FreeBSD armv6


MikaelUrankar:
Hi,

We are seeing this error on FreeBSD armv6 with clang 3.8.0:
```
[ 21%] Building C object interface/khronos/CMakeFiles/EGL_static.dir/vg/vg_int_mat3x3.c.o
/usr/obj/portswork/usr/ports/misc/raspberrypi-userland/work/userland-fcbf8a3/interface/mmal/util/mmal_param_convert.c:47:7: error: variable 'i' is incremented both in the loop header and in the loop body [-Werror,-Wfor-loop-analysis]
      i++;
      ^
/usr/obj/portswork/usr/ports/misc/raspberrypi-userland/work/userland-fcbf8a3/interface/mmal/util/mmal_param_convert.c:40:25: note: incremented here
   for (i=0; i<n_pairs; i++)
                        ^
1 error generated.
```

The issue is described in details here:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=207097

6by9:
Yes, looks like a genuine bug. Feel free to create a pull request!

popcornmix:
This is now fixed.

PR #292, opened 2016-02-15
properly update loop indexes once

MikaelUrankar:
Fix clang warning about changing the loop variable in both loop
body and header

Fix issue #291

### DIFF
--- a/interface/mmal/util/mmal_param_convert.c
+++ b/interface/mmal/util/mmal_param_convert.c
@@ -37,14 +37,13 @@ typedef struct string_pair_t
 static MMAL_STATUS_T parse_enum(int *dest, string_pair_t *pairs, size_t n_pairs, const char *str)
 {
    size_t i;
-   for (i=0; i<n_pairs; i++)
+   for (i=0; i<n_pairs; i+=2)
    {
       if (vcos_strcasecmp(str, pairs[i].string) == 0)
       {
          *dest = pairs[i].value;
          return MMAL_SUCCESS;
       }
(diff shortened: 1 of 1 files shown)


popcornmix:
While this avoids the compiler complaint, I suspect ``i`` should just be incremented by 1 each iteration.
@6by9 do you agree?

6by9:
I'd agree, ```i``` should only be incremented by 1 within the ```for``` line

JamesH65:
Agreed, but how has this code ever worked? Is it never called, or is it by
accident that it mostly works?

6by9:
Almost never used.
mmal_parse_video_codec and mmal_parse_video_size are I think the only two users, and I think those only got used by mmalcam and mmalplay. 
I did check this morning and this bug was present on ISYS too, never altered since initial commit by my old boss :-)

JamesH65:
Clearly thoroughly unit tested at the time then....

popcornmix:
I've removed the i++ from the body of the function on the internal tree. This will appear in userland next time I rebuild.

6by9:
> Clearly thoroughly unit tested at the time then....

I think I might refrain from commenting. It was back in Oct 2011, so before code reviews were enforced.

I've just felt nostalgic seeing 
```"[git-p4: depot-paths = "//software/.....": change = 292021]```
at the bottom of the commit. I'll stop now before I start telling stories about when I were a lad...

popcornmix:
I've pushed a version of this - thanks for reporting.

PR #293, opened 2016-02-18
loop is already incrementing the index

kraj:
It seems a typo, after this the index will be icremented
by 2 in the loop, is that correct ?


kraj:
already fixed by #292

PR #294, opened 2016-02-18
Allow to override VMCS_INSTALL_PREFIX

ggardet:
Allow to override VMCS_INSTALL_PREFIX to be able to install to a user defined place


pelwell:
Why not give the Android builders the option of choosing a different location?

ggardet:
This is not for Android but for Linux distributions (such as openSUSE).
/opt/vc is not a standard location. So, we install it in /usr instead. It would be nice if we could set it at compile time. 
Note that there is no default behavior change with this patch.

Issue #133, opened 2013-12-28
Raspistill buffer flush with standard output


Pol-on:
When images are saved to file there is no problem but when the images are sent to standard output using "-o -" sometimes part of an image is missing. To reproduce this you can use
raspistill -w 640 -h 480 -n -t 0 -s -o - | less > test.jpg
In smaller images the problem is better visible because bigger part of the image is missing. 
If the problem happens no more images can be save. 
This is probably caused by a buffer. Not all data go to the output and corrupt the next image. 
I think the solution is relatively simple. The buffer needs to be flashed when the output is finished. 
I use the "-o -" option to get the images to my program. It is faster than saving image to disk and reading them.

Ruffio:
@Pol-on does the fix work?

Pol-on:
blockquote {padding-left: 1ex; margin: 0px 0px 0px 0.8ex; border-left: #cccccc 1px solid;} p {margin: 0px;padding: 0px;} 
Yes, it does.
It was very simple.
Best wishes
Adam
 
Dnia Pi±tek, 26 Czerwca 2015 23:48 Rasmus Christiansen <notifications@github.com> napisa³(a)

@Pol-on does the fix work?

PR #297, opened 2016-03-05
raspiyuv: Flush stdout to ensure image is fully sent

kfitzgerald:
This PR adds a manual flush after writing the image data to stdout.

Without the flush, using signals to stream image data over stdout would result in the missing data, if the image didn't completely fill the buffer.

For validation, a 640x480 YUV image will consist of the following number of bytes:
 * Y: 640 x 480 = 307200 bytes
 * U: 640 x 480 ÷ 4 = 76800 bytes
 * V: 640 x 480 ÷ 4 = 76800 bytes
 * Sum: **460800 bytes**.

## Example

Prior to the patch, this is what happens when you pipe the image data over standard out and wait for SIGUSR1 to take a shot.

```sh
raspiyuv -o - -w 640 -h 480  --nopreview --signal | wc
  18134    2498  458752
```

Notice how `458752` is exactly `7 x 65536`. *It's missing the remaining `2048` bytes*, which doesn't automatically trigger a flush.

And after the patch:

```sh
$ raspiyuv -o - -w 640 -h 480  --nopreview --signal | wc
  17043    2602  460800
```

For both examples, triggering the shot in a separate shell:
```sh
kill -SIGUSR1 $(ps -C raspiyuv --no-headers -o "pid")
sleep 2
kill $(ps -C raspiyuv --no-headers -o "pid")
```

Please let me know if there is anything else you need from me. 

Thanks!
-Kevin

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStillYUV.c
+++ b/host_applications/linux/apps/raspicam/RaspiStillYUV.c
@@ -1376,6 +1376,10 @@ int main(int argc, const char **argv)
                {
                   rename_file(&state, output_file, final_filename, use_filename, frame);
                }
+               else
+               {
+                  fflush(output_file);
+               }
             }
 
             if (use_filename)


JamesH65:
Looks good to me. Thanks for the patch.

popcornmix:
@6by9 happy with this?

6by9:
Yup, looks reasonable.
It seems to be the same thing as already present in RasiStill https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiStill.c#L1992 from #133.

PR #298, opened 2016-03-14
Adjust include paths for Linux/BSD.

ED6E0F17:
Include pthreads path directly on __UNIX__ . Tested that Userland still builds on Raspbian, but may  have impact on other platforms e.g. Android.

### DIFF
--- a/interface/vcos/pthreads/vcos_platform.h
+++ b/interface/vcos/pthreads/vcos_platform.h
@@ -29,7 +29,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 VideoCore OS Abstraction Layer - pthreads types
 =============================================================================*/
 
-/* Do not include this file directly - instead include it via vcos.h */
+/* DO NOT include this file directly - instead include it via vcos.h */
 
 /** @file
   *
--- a/interface/vcos/vcos.h
+++ b/interface/vcos/vcos.h
@@ -113,7 +113,12 @@ VideoCore OS Abstraction Layer - public header file
 
 #include "interface/vcos/vcos_assert.h"
 #include "vcos_types.h"
+
+#if defined(__unix__) && !defined(__ANDROID__)
+#include "interface/vcos/pthreads/vcos_platform.h"
+#else
 #include "vcos_platform.h"
+#endif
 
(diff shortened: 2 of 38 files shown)


popcornmix:
Have you got a link that clarifies when ``__UNIX__`` is defined?
We'd like to sure it is defined for all linux/pthread platforms, but not for other supported platforms (android is the worry).

ED6E0F17:
I would expect that Android IS Unix, so I have guarded against that now, although I would have to use the Linux pthreads path myself if I wanted to build Userland for Brillo.

I have also removed the extra Number Sign and squashed the commits.

__UNIX__ seems to be defined for GCC and Clang for all BSD/Linux except OSX :
http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system

popcornmix:
gcc with no options on both Ubuntu native, Ubuntu cross compiling for arm and raspbian native define ``__unix__`` and ``__linux__`` but not ``__UNIX__`` or ``__LINUX__``
As such I can't cross compile omxplayer for example with this PR.

popcornmix:
Try http://paste.ubuntu.com/15384596/ with ``gcc test.c``

popcornmix:
Also this seems to only be a partial solution.
If I remove include paths from omxplayer it just fails with other includes like:
https://github.com/raspberrypi/userland/blob/master/interface/vmcs_host/vcgencmd.h#L33

ED6E0F17:
It does look bad when I am using uppercase __UNIX__ and link to a page that shows all lowercase __unix__ without noticing the difference.

"vchost_config.h" is in another path; I will see what I can do for that.

ED6E0F17:
Fixed case for __unix__ and checked that __ANDROID__ is uppercase and that the  Android toolchain defines both __unix__ and __linux__

Checked that Userland still builds on Raspbian, after fixing another mistake.

Applied the same logic when including vchost_config.h

popcornmix:
I've been using this locally for some weeks. I've made a couple of minor tweaks and have now pushed it out. Hopefully it will be fine. Thanks.

ED6E0F17:
Thanks. This only affects a handful of users every year, but it should close a few issues in the bug tracker.

PR #300, opened 2016-03-24
Fix errors due to ignored return code

kraj:
Clang warns/errors about not checking the return
values from chdir() and fread(), here we check
for chdir return code and report back using error()
API, and second fix to fread is to let compiler
ignore the value via assigning it to an unused variable


pelwell:
Thanks - I've folded this into another commit with a few minor tweaks.

PR #301, opened 2016-03-29
Use of glob.h breaks Android builds.

ED6E0F17:
glob is not used here, so glob.h need not be included.

Also, sprintf is obsolete and throws compiler warnings no matter how obviously safe this one is. This is not the only example in the code base, but there is no need to make things worse.


pelwell:
Thanks - I've folded this into another commit with a few minor tweaks.

PR #304, opened 2016-04-02
build shared library for vchostif

kraj:
Fixes #149

Signed-off-by: Khem Raj raj.khem@gmail.com

### DIFF
--- a/interface/vmcs_host/CMakeLists.txt
+++ b/interface/vmcs_host/CMakeLists.txt
@@ -9,7 +9,7 @@ add_definitions(-fno-strict-aliasing)
 
 include_directories(${VMCS_TARGET}/vcfiled)
 
-add_library(vchostif
+add_library(vchostif SHARED
             ${VMCS_TARGET}/vcfilesys.c ${VMCS_TARGET}/vcmisc.c
             vc_vchi_gencmd.c vc_vchi_filesys.c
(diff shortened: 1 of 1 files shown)


pelwell:
What does it fix? https://github.com/raspberrypi/userland/pull/149 doesn't look relevant.

kraj:
Actually Its issue from firmware repo..
see https://github.com/raspberrypi/firmware/issues/149

Ruffio:
@popcornmix is this good to go?

popcornmix:
There are a couple of gotchas. There is a new file ``libvchostif.so`` and a missing file ``libvchostif.a``
Depending on how the library is updated you may get this error:
```
error while loading shared libraries: libvchostif.so: cannot open shared object file: No such file or directory
```
until you next run ``sudo ldconfig``. rpi-update and apt-get should handle that, but there may be distributions that just copy files (including using a fixed list that doesn't include ``libvchostif.so``).
Makefiles that hard code in the libvchostif.a library will break and need updating.

It's a bit hard to tell how many people (if any) this will inconvenience.
One option is to push it and see what the response it (it can always be reverted).

@pelwell @6by9 any views?

popcornmix:
Yes, just checked and kodi doesn't build without the addition of (presumably) -lvchostif, so there definitely will be inconvenience for some with this change.

6by9:
I've no real views on this one. Defer to others.

PR #308, opened 2016-04-16
Fix unixtime '1900' typo in raspistill --help

truedays:


popcornmix:
Looks okay to me. @6by9 @JamesH65 any objections?

6by9:
Nothing too radical, so I'm happy.
(James is going to be slightly distracted with other things at the moment)

PR #310, opened 2016-04-28
khronos: Use vchiq_get_client_id to determine global PID.

lorenzo-stoakes:
VCHIQ under the covers uses PID as a unique identifier for each 'service' that
ferries messages back and forth from the GPU.

This becomes an issue when PID namespacing is in effect, for example when
running a containerised application, as duplicate PIDs then become possible
potentially resulting in clashes between different VCHIQ clients.

Additionally, when running in a PID namespace, the kernel part of the VCHIQ code
will use a different identifier than that used by client code, which results in
VCHIQ failing to work correctly _at all_ when one is in effect.

Fortunately the globally unique PID is available as a 'client ID' via an
existing ioctl - GET_CLIENT_ID. By using this ioctl rather than obtaining
process PID we avoid the issue entirely.

### DIFF
--- a/interface/khronos/common/khrn_client.c
+++ b/interface/khronos/common/khrn_client.c
@@ -372,7 +372,7 @@ static uint32_t convert_gltype(EGL_CONTEXT_TYPE_T type)
 
 void client_send_make_current(CLIENT_THREAD_STATE_T *thread)
 {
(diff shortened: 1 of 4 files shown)


lorenzo-stoakes:
Whoops this is an early version did not mean to actually submit this just yet :)

PR #312, opened 2016-04-30
Fix compiler warnings

luked99:
This fixes a few small compiler warnings in userland.

The most controversial is removing the usbdk code. As far as I can tell it's not actually used. A search on google doesn't throw up anyone asking about it, but it generates a ton of compiler warnings.


interface/khronos/common/linux/khrn_client_platform_linux.c
```diff
@@ -271,7 +271,7 @@ static KHRN_IMAGE_FORMAT_T convert_format(uint32_t format)
       case EGL_PIXEL_FORMAT_XRGB_8888_BRCM:     return XBGR_8888;
       case EGL_PIXEL_FORMAT_RGB_565_BRCM:       return RGB_565;
       case EGL_PIXEL_FORMAT_A_8_BRCM:           return A_8;
-      default:                                  vcos_verify(0); return (KHRN_IMAGE_FORMAT_T)0;
+      default:                                  vcos_assert(0); return (KHRN_IMAGE_FORMAT_T)0;
```
popcornmix: Was this one deliberately made an assert, rather than a verify cast to void?

interface/khronos/glxx/glxx_client.c
```diff
@@ -2989,7 +2989,7 @@ GL_API GLboolean GL_APIENTRY glIsEnabled (GLenum cap)
       {
          GLboolean temp = 0;
          GLuint count = (GLuint) get_boolean_internal_11(thread, cap, &temp);
-         UNUSED_NDEBUG(count);
+         (void)count;
```
popcornmix: Is this change needed given the updated UNUSED_NDEBUG definition?

ED6E0F17:
aspect_ratio_sd_str() is NOT unused, it just doesn`t get called when it should because of a bug in this line calling the wrong routine: 

    ./host_applications/linux/apps/tvservice/tvservice.c:423:  
    tmp = status_sprintf(mode_str, MAX_STATUS_STR_LENGTH, &offset, " %s", aspect_ratio_str(tvstate->display.sdtv.display_options.aspect));

luked99:
...although aspect_ratio_str() appears to return the exact same result as
aspect_ratio_sd_str() for all valid SD aspect ratios....

luked99:
Another try at this. This just removes the usbdk code, and I'll then take a look at the other changes later.

popcornmix:
usbdk is not used, and firmware and userland still build with this change so merging.
Thanks.

PR #321, opened 2016-06-18
RaspiStill: Update annotation before capture

### LINKED ISSUE #319: raspistill --timelapse + --annotate puts the same timestamp on every frame
Using raspistill --annotate 12 --timelapse <interval>, every frame is annotated with an identical timestamp - the time that raspistill was started. It should use the time that the frame was taken.

(libraspberrypi-bin 1.20160523-1)

ethanol100:
Fixes #319


popcornmix:
@6by9 is this okay with you?

6by9:
Yes, looks fine.

PR #322, opened 2016-06-20
Update raspicam helptext

lurch:
Valid --drc options are mentioned in the Notes (just like the valid options for --exposure, --awb, --imxfx and --metering)


JamesH65:
Looks good to me.

PR #324, opened 2016-07-23
Update RaspiStill.c

choury:
This should be a typo


popcornmix:
Looks correct - thanks.

PR #325, opened 2016-07-27
openvg: vguPolygon incorrect coordinate pointer increment

paeryn:
The pointer to the start of the next coordinate to send was incorrectly incremented (by 1*count rather than the 2*count sent).

### DIFF
--- a/interface/khronos/vg/vg_client.c
+++ b/interface/khronos/vg/vg_client.c
@@ -5122,7 +5122,7 @@ VGU_API_CALL VGUErrorCode VGU_API_ENTRY vguPolygon(
             rpc_send_ctrl_end(thread);
 
             ps_count -= chunk_ps_count;
-            ps += chunk_ps_count;
+            ps += 2 * chunk_ps_count;
             first = false;
          }
       }


JamesH65:
Just for my own understanding, this change looks completely valid, and yet this issue has never been seen before? Surely the polygon function must have been tested at some point? This issue would have produced completely garbled results. Or is this code path underused?

paeryn:
Probably underused, I only noticed it on large polygons (the first incorrect vertex happens at around the 245th vertex). I flagged the error up here https://github.com/raspberrypi/firmware/issues/633 and popcornmix suggested the fix. Since changing it my test program draws correctly.

JamesH65:
Ah, makes sense, it's 'chunking' the data, so only goes wrong when you have more than one chunk, which would be around 245 vertices. Cool, thanks.

popcornmix:
I've already got this fix in the internal repo - it will appear in /opt/vc libs and userland on next firmware update.

popcornmix:
Okay this is now in latest rpi-update and in userland tree.
Thanks for report.

Issue #170, opened 2014-05-09
raspiyuv: request for output Y values only.


chewep:
Is it possible to output Y values only and discarding U and V values?

Ruffio:
@chewep is this still an issue?

Ruffio:
@chewep is this still an issue?

waveform80:
Not even necessary to hack raspiyuv for this - simply pipe the output through the standard `head` command to chop off the UV tail and you should be good to go. Just calculate the number of bytes you need (bearing in mind the camera rounding horizontal resolution up to 32, and the vertical up to 16).

For example, with a V1 camera module capturing at 2592x1944 the amount of data output by raspiyuv will be 2592 * 1952 * 1.5 (1952 because of the vertical rounding, 1.5 because YUV420 format includes Y values at full resolution and U and V values at quarter resolution each). We only want the first 2592 * 1952 = 5059584 bytes of information, which the following command line should achieve:

```
$ raspiyuv -o - | head -c $((2592*1952)) > output.yuv
```

Just adjust the `-c` value according to your resolution and you should be good to go.

6by9:
You can use head, but that won't work for timelapse sensibly.

Boredom, so just making up a PR. raspividyuv also appears to be missing support for RGB mode, so stuffing that in at the same time.

6by9:
Patches pushed to my tree, but totally untested beyond "they build", so need to do that before creating a PR.

PR #330, opened 2016-08-15
Fix for issue #170, and raspividyuv supporting rgb

6by9:


6by9:
Rebased and pushed again. It builds and the conflicts were minimal, so should be safe to push even though I haven't actually tested it :-o

PR #331, opened 2016-08-26
Do not return random data

plusky:
This would solve issues  with openSUSE packagign - automatic check complaining:
```
I: Program returns random data in a function
E: raspberrypi-userland no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/raspberrypi-userland-2016.08.17/interface/vchiq_arm/vchiq_lib.c:164

```


popcornmix:
Looks like that should actually be:
```   return vchiq_initialise_fd(pinstance, -1);```

I've fixed our internal repo, it will appear in userland on next update.

plusky:
Thanks

PR #332, opened 2016-08-30
Add name to anonymous structs in RaspiCamControl.h

jogu:
This is useful as it allows applications to pass around struct
RASPICAM_CAMERA_PARAMETERS * parameters without the need to include
RaspiCamControl.h and all it's dependencies.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.h
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.h
@@ -105,20 +105,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
 // There isn't actually a MMAL structure for the following, so make one
-typedef struct
+typedef struct mmal_param_colourfx_s
 {
    int enable;       /// Turn colourFX on or off
    int u,v;          /// U and V to use
 } MMAL_PARAM_COLOURFX_T;
 
-typedef struct
+typedef struct mmal_param_thumbnail_config_s
 {
    int enable;
    int width,height;
    int quality;
 } MMAL_PARAM_THUMBNAIL_CONFIG_T;
 
-typedef struct
+typedef struct param_float_rect_s
 {
    double x;
    double y;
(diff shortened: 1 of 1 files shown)


jogu:
Hi there,

I'm trying to merge the mmal enabled fork of motion back into the upstream ( https://motion-project.github.io ).

The current fork has modified versions of some of the userland files included it in, just to add a name to one of the anonymous structs, so I'm hoping to be able to merge that change so that we can use your unmodified code.

(btw: is there any general guidance or plans around how third-party camera apps should get access to mmal? Currently the build instructions for the fork start with cloning userland.git; I was expecting the files like say mmal_parameters_camera.h etc might be available in a dev dpkg installable via apt-get, but I've not yet found anything like that.)

Thanks

Joseph

jogu:
Ah, ignore my question - I found the libraspberrypi-dev package :) - I shall update the fork to use that instead. Any advice on best way to do things is still very welcome though.

JamesH65:
Patch looks fine to me - still required?

jogu:
Thanks @JamesH65 - yes, this patch is still required. (Apologies for any confusion, I'm fighting multiple battles at the same time to try and get the forked knocked into a good shape!)

pelwell:
I'd prefer struct names to be lowercase, but otherwise I'm happy with the patch.

popcornmix:
Yes, the general coding standard would be:
```typedef struct mmal_param_colourfx_s```

jogu:
Thanks @pelwell & @popcornmix 

I've pushed a new update that changes them to lowercase with the _s suffix; apologies, was unaware of the coding standard.

popcornmix:
@6by9 happy?

6by9:
Happy enough.

(I'm currently missing the gain that it gives as this is still inside RaspiCamControl.h. Or are you just forward defining struct raspicam_camera_parameters_s inside other headers, and only those files that access the contents include RaspiCamControl.h to provide the implementation?)

JamesH65:
My assumption is the latter. But you know what they say about assumptions.

jogu:
@6by9 Exactly the gain you say :-) It's a relatively minor gain, but does make things easier for us.

PR #334, opened 2016-09-26
RaspiVid: Make information in comments correct and fix data type

adamheinrich:
Comments should reflect the actual situation to prevent confusion. Correct data type prevents compiler from complaining.


adamheinrich:
I've fixed the commit message, thanks!

Issue #188, opened 2014-07-22
Compile warnings need followup. 


rewolff:
A few days ago I tried compiling this repository. Waiting for the compile to finish I was bored and investigated one of the warnings. That pointed at a line that was CLEARLY a bug.... 

An assignment `
    x = <someconstant>;`
was written as `
    x == <someconstant>;`

(So sorry I don't remember where exactly this was, but investigating more than ONE of the warnings needs to be done anyway...) 

-> TODO clean up all the warnings. 
The reward is that you'll find and fix at least one bug in doing so....

Ruffio:
@rewolff @popcornmix is this still an issue?

Ruffio:
@popcornmix has this been fixed?

rewolff:
The "bug" that the compiler noticed has been fixed. Still a lot of compile errors remain. 
I've fixed a few:  http://prive.bitwizard.nl/compile_fixes.patch

There are  a bunch of: 
> warning: implicit declaration of function ‘sbrk’ 
warnings that I haven't been able to fix, but my attempts remain in the patch. 

The manual for sbreak says it is declared in unistd.h, and including it doesn't help. Declaring it perhaps a bit sloppy makes the compiler give errors because of the conflicting definitions. 

I have changed a "notused" macro that was used to indicate that a variable was not-used (when not debugging). This seems to work. I'm not sure it will work for all compilers that can be used. 

Oh. I'm compiling natively on a recent raspbian.

luked99:
The sbrk() warning can be fixed with https://github.com/luked99/userland-1/commit/627eaebe50a26e28be365aca03efb932a25228cf.

The UNUSED() macros can be fixed with https://github.com/luked99/userland-1/commit/9f02540a439be25a312cafe2c81ef4104f9dff87.

I need to try these out on a pi; then I'll create a pull request (they should be OK but you can never be too sure).

JamesH65:
I believe this may now be fixed with the merges above?

This issue will be closed within 30 days unless further interactions are posted. If you wish this issue to remain open, please add a comment. A closed issue may be reopened if requested.

JamesH65:
Closing due to lack of activity. Please request to be reopened if you feel this issue is still relevant.

PR #335, opened 2016-09-26
Fix definition of UNUSED(x)

luked99:
Make UNUSED(x) explicitly cast (x) to void. This fixes a number
of compiler warnings.

See also #188.

### DIFF
--- a/interface/khronos/common/khrn_int_common.h
+++ b/interface/khronos/common/khrn_int_common.h
@@ -162,14 +162,10 @@ typedef unsigned long long uint64_t;
 #ifdef _MSC_VER
    #define UNUSED(X) X
 #else
-   #define UNUSED(X)
+   #define UNUSED(X) (void)(X)
 #endif
 
-#ifdef NDEBUG
-   #define UNUSED_NDEBUG(X) UNUSED(X)
-#else
-   #define UNUSED_NDEBUG(X)
-#endif
+#define UNUSED_NDEBUG(X) UNUSED(X)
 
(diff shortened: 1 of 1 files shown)


interface/khronos/common/khrn_int_common.h
```diff
@@ -162,13 +162,13 @@ typedef unsigned long long uint64_t;
...
 #else
-   #define UNUSED(X)
+   #define UNUSED(X) (void)(X)
 #endif
 
 #ifdef NDEBUG
    #define UNUSED_NDEBUG(X) UNUSED(X)
 #else
-   #define UNUSED_NDEBUG(X)
+   #define UNUSED_NDEBUG(X) UNUSED(X)
```
kukabu: hm... then why ifdef? line 169 and 171 identical now

luked99: Good suggestion.

luked99: I've removed the ifdef.

popcornmix:
This reduces warnings from 52 to 20.
There are still a couple of warnings of the form:
```
/home/dc4/projects/vc4/interface/khronos/common/linux/khrn_client_platform_linux.c: In function 'convert_format':
/home/dc4/projects/vc4/interface/khronos/common/linux/khrn_client_platform_linux.c:261:16: warning: statement with no effect [-Wunused-value]
       default:                                  vcos_verify(0); return (KHRN_IMAGE_FORMAT_T)0;
                ^
```
But that can be resolved at a later date.

Issue #316, opened 2016-06-02
warning: implicit declaration of function 'sbrk' and 'strdup'


fordfrog:
while installing latest (20160531) raspberrypi-userland on gentoo, i got these QA warnings:

```
 * QA Notice: Package triggers severe warnings which indicate that it
 *            may exhibit random runtime failures.
 * /var/tmp/portage/media-libs/raspberrypi-userland-0_pre20160531/work/raspberrypi-userland-ed59ee1/interface/vcos/pthreads/vcos_platform.h:777:4: warning: implicit declaration of function ‘sbrk’ [-Wimplicit-function-declaration]
 * /var/tmp/portage/media-libs/raspberrypi-userland-0_pre20160531/work/raspberrypi-userland-ed59ee1/interface/vcos/pthreads/vcos_platform.h:818:4: warning: implicit declaration of function ‘strdup’ [-Wimplicit-function-declaration]
```

i don't know whether it is intended or it should be fixed so can someone who knows this stuff please clarify it?

fordfrog:
i found this info: http://stackoverflow.com/questions/8440816/warning-implicit-declaration-of-function

so it means that these two functions are used before they are declared (either in header file or before they are first used).

popcornmix:
``sbrk`` should be declared in ``<unistd.h>`` which is being included. 
``strdup`` should be declared in ``<string.h>`` which is being included.
I'm not sure if the gentoo compiler doesn't support these?

I don't think ``sbrk`` or ``strdup`` are typically used by userland (mostly for test functions), and the lack of a declaration for a function with simple types is normally not fatal, so it's probably not a big issue.

@6by9 @pelwell any thoughts?

pelwell:
strdup is part of POSIX so ought to be included. sbrk has been removed, so perhaps that failure is more understandable. Which compiler (including version) are you using?

ED6E0F17:
I have seen this error previously, and now that I have checked "string,h" on wheezy I understand why.
strdup is not defined, strdupa is.

ED6E0F17:
It look as though "_GNU_SOURCE" needs to be defined to get "USE_XOPEN_EXTENDED", and then define sbrk and strdup. "_GNU_SOURCE" is defined in some places, but not everywhere that it is needed.

fordfrog:
@pelwell gcc 4.9.3

ED6E0F17:
[...]

PR #336, opened 2016-09-27
Remove vcos_get_free_mem()

luked99:
This function isn't used anywhere, and causes compiler warnings
due to its use of sbrk() without the relevant #defines.

The "pthreads" version of this doesn't do anything sensible: on
a linux system it should look at something like /proc/vmstat,
but that's not what it does. So additionally remove it to avoid
confusion.

See issue #316.

### DIFF
--- a/interface/vcos/pthreads/vcos_platform.h
+++ b/interface/vcos/pthreads/vcos_platform.h
@@ -674,11 +674,11 @@ void vcos_atomic_flags_delete(VCOS_ATOMIC_FLAGS_T *atomic_flags)
 
 #if defined(linux) || defined(_HAVE_SBRK)
 
-/* not exactly the free memory, but a measure of it */
+/* Returns invalid result, do not use */
 
 VCOS_INLINE_IMPL
-unsigned long vcos_get_free_mem(void) {
-   return (unsigned long)sbrk(0);
+unsigned long VCOS_DEPRECATED("returns invalid result") vcos_get_free_mem(void) {
+   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


6by9:
Checking full firmware source, it is used directly in RUT, audio tests, ARTS performance, and ILTest.
Admittedly those haven't been used in a fair while (self hosted was broken last time I tried), but it's probably safer to drop the pthreads version to being "return 0;" rather than removing the whole function and prototype.

popcornmix:
Agreed - we don't currently build any of the test apps that call vcos_get_free_mem, but they are in the tree and could be built in theory.
Considering it deprecated and allowing "return 0" seems a reasonable option.
(if removing it a few other prototypes would need removing anyway).

luked99:
Better still, make it with __attribute__((deprecated)) ?

popcornmix:
Do we have a portable way of doing that? Would need to compile with hcvc, msvc, gcc etc.

luked99:
Perhaps something like:


    #ifdef __GNUC__
    #define VCOS_DEPRECATED __attribute__((deprecated))
    #else
    #define VCOS_DEPRECATED
    #endif

Then put VCOS_DEPRECATED in the pthreads header file but not the common one (which may or may not be accepted by the compiler).

Totally untested.

luked99:
877efd2a55516e76247b54a23d9a996092fe54a0 deprecates the pthreads version of vcos_get_free_mem().

popcornmix:
Thanks.

PR #338, opened 2016-09-28
RaspiVid: Make open_filename() unified

adamheinrich:
Replace three separate functions `open_filename`, `open_imv_filename` and `open_pts_filename` by a single `open_filename`.

This change:
- Allows the user to stream motion vectors and timestamps to socket as well
- Removes repetitive code

Also rename `pts_file` to `pts_filename` to avoid confusion as other variables are called `filename` and `imv_filename`.


popcornmix:
Looks good to me.
@6by9 ?

JamesH65:
Looks fine to me but not tested. I wonder why that hasn't already been done...Must pay more attention.

6by9:
Again not tested, but looks good.

PR #340, opened 2016-09-29
buildme: avoid shell expansion of tr expression

luked99:
I found out by accident that buildme doesn't do quite enough escaping of the "tr" command's arguments. It was quite puzzling....


popcornmix:
Yep!

PR #341, opened 2016-09-29
Warnings/vcos verify

luked99:
RFC: remove a few warnings around vcos_verify().

Vcos_verify() was originally intended to be used like this:

    if (!vcos_verify(condition_expected_to_be_true))
        handle_unexpected_error_somehow();

The idea is that in normal use vcos_verify just evaluates its argument, but if you are trying to debug a problem it can be flipped via a build-time switch to generate a breakpoint and drop to a debugger.

The compiler is now warning that the rvalue is ignored in a few places where it is used outside a conditional stmnt.

It is not correct to blithely do s/vcos_verify/vcos_assert/g as behaviour will change.

One fix is to cast all the result to (void). I have actually chosen to switch one instance to a vcos_assert() because the code there looks like it should be an assert, and leave the others alone (but cast to void to squelch the warning).


popcornmix:
Let me run with this for a couple of days to make sure that assert doesn't occur (I doubt it does).

popcornmix:
Not seen any issues with this. Thanks.

PR #342, opened 2016-09-29
RaspiVid: Add raw (YUV420, RGB or grayscale) video output

adamheinrich:
The raw video output is useful for image processing applications where both motion vectors from encoder and image data are needed. This pull request adds a new `--raw` option which inserts a video splitter component between camera's video output and encoder's input, which allows us to set callback to capture raw video data into a file.

Another option `--raw-format` is introduced which allows us to choose YUV420, RGB or grayscale format.

Parts of the code (video output format configuration and `camera_buffer_callback()`) come from RaspiYUV.c.

The splitter component is created only if the `--raw` option is specified.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -43,6 +43,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * components, but we do need to handle buffers from the encoder, which
  * are simply written straight to the file in the requisite buffer callback.
  *
+ * If raw option is selected, a video splitter component is connected between
+ * camera and preview. This allows us to set up callback for raw camera data
+ * (in YUV420 or RGB format) which might be useful for further image processing.
+ *
  * We use the RaspiCamControl code to handle the specific camera settings.
  * We use the RaspiPreview code to handle the (generic) preview window
  */
@@ -84,6 +88,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define MMAL_CAMERA_PREVIEW_PORT 0
 #define MMAL_CAMERA_VIDEO_PORT 1
 #define MMAL_CAMERA_CAPTURE_PORT 2
+
+// Port configuration for the splitter component
+#define SPLITTER_OUTPUT_PORT 0
+#define SPLITTER_PREVIEW_PORT 1
 
 // Video format information
 // 0 implies variable
@@ -141,9 +149,18 @@ typedef struct
    char  header_bytes[29];
    int  header_wptr;
    FILE *imv_file_handle;               /// File handle to write inline motion vectors to.
+   FILE *raw_file_handle;               /// File handle to write raw data to.
    int  flush_buffers;
    FILE *pts_file_handle;               /// File timestamps
 } PORT_USERDATA;
+
+/** Possible raw output formats
+ */
+typedef enum {
+   RAW_OUTPUT_FMT_YUV = 1,
+   RAW_OUTPUT_FMT_RGB,
+   RAW_OUTPUT_FMT_GRAY,
+} RAW_OUTPUT_FMT;
 
 /** Structure containing all state information for the current run
  */
@@ -181,10 +198,13 @@ struct RASPIVID_STATE_S
    RASPICAM_CAMERA_PARAMETERS camera_parameters; /// Camera setup parameters
 
    MMAL_COMPONENT_T *camera_component;    /// Pointer to the camera component
+   MMAL_COMPONENT_T *splitter_component;  /// Pointer to the splitter component
    MMAL_COMPONENT_T *encoder_component;   /// Pointer to the encoder component
-   MMAL_CONNECTION_T *preview_connection; /// Pointer to the connection from camera to preview
+   MMAL_CONNECTION_T *preview_connection; /// Pointer to the connection from camera or splitter to preview
+   MMAL_CONNECTION_T *splitter_connection;/// Pointer to the connection from camera to splitter
    MMAL_CONNECTION_T *encoder_connection; /// Pointer to the connection from camera to encoder
 
+   MMAL_POOL_T *splitter_pool; /// Pointer to the pool of buffers used by splitter output port 0
    MMAL_POOL_T *encoder_pool; /// Pointer to the pool of buffers used by encoder output port
 
    PORT_USERDATA callback_data;        /// Used to move data to the encoder callback
@@ -194,6 +214,9 @@ struct RASPIVID_STATE_S
 
    int inlineMotionVectors;             /// Encoder outputs inline Motion Vectors
    char *imv_filename;                  /// filename of inline Motion Vectors output
+   int raw_output;                      /// Output raw video from camera as well
+   RAW_OUTPUT_FMT raw_output_fmt;       /// The raw video format
+   char *raw_filename;                  /// Filename for raw video output
    int cameraNum;                       /// Camera number
    int settings;                        /// Request settings from the camera
    int sensor_mode;			            /// Sensor mode. 0=auto. Check docs/forum for modes selected by other values.
@@ -247,6 +270,14 @@ static XREF_T  intra_refresh_map[] =
 
 static int intra_refresh_map_size = sizeof(intra_refresh_map) / sizeof(intra_refresh_map[0]);
 
+static XREF_T  raw_output_fmt_map[] =
+{
+   {"yuv",  RAW_OUTPUT_FMT_YUV},
+   {"rgb",  RAW_OUTPUT_FMT_RGB},
+   {"gray", RAW_OUTPUT_FMT_GRAY},
+};
+
+static int raw_output_fmt_map_size = sizeof(raw_output_fmt_map) / sizeof(raw_output_fmt_map[0]);
 
 static void display_valid_parameters(char *app_name);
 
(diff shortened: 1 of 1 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1448,7 +1594,30 @@ static MMAL_STATUS_T create_camera_component(RASPIVID_STATE *state)
    // Set the encode format on the video  port
 
    format = video_port->format;
-   format->encoding_variant = MMAL_ENCODING_I420;
+
+   if (state->raw_output)
+   {
+      switch (state->raw_output_fmt)
```
6by9: Unless I've missed something, this is not going to be great.
Video_splitter should be able to do the format conversions (if not then that needs fixing), so significantly better to keep this connection as ENCODING_OPAQUE, and do the conversion on the one output of video_splitter. Otherwise in the RGB case camera will be converting from YUV to RGB, and video_encode will be converting back again as it requires YUV.

I would actually recommend putting the splitter on the preview output (camera->output[0]) and leave output[1] going to the encoder as ENCODING_OPAQUE. Due to various historic reasons, camera->output[1] is optimised for video_encode and provides extra data that makes the encode more efficient. video_splitter doesn't support those optimisations, so inserting it between the two breaks part of the encode path.

adamheinrich: Thank you for the information! I'll try to configure the splitter to pass ENCODING_OPAQUE data from camera to encoder and to do format conversion on output port.

6by9: Yes, still not fully optimal, but much lower overhead compared to the YUV to RGB conversions you did have.

If you do switch to output[0], then both sets of buffers should be timestamped, so you should be able to tally them. There's no difference at all in the actual image data sent from the two ports - it was more to match the OpenMaxIL spec where the camera component had 2 ports, one for preview and one for capture (extended to 3 ports in a change that I don't believe ever made it into an updated version of the spec).

You'd have an issue even if using output[1], as motion vectors are post video_encode and that has an input FIFO and processing latency. There is unlikely to be any significant difference between that and the processing delay if you use output[0].

adamheinrich: I've changed the commit again. The video splitter is now connected between camera's output[0] and preview's input.

6by9:
Sorry, I haven't totally forgotten these, but not had a chance to review. Will try to do so in the next day or so.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -806,6 +844,40 @@ static int parse_cmdline(int argc, const char **argv, RASPIVID_STATE *state)
...
+         break;
+      }
+
+      case CommandRawFormat:
+      {
+         state->raw_output_fmt = raspicli_map_xref(argv[i + 1], raw_output_fmt_map, raw_output_fmt_map_size);
+         i++;
+
+         if (state->raw_output_fmt == -1)
+            state->raw_output_fmt = RAW_OUTPUT_FMT_YUV;
```
6by9: So if they enter "--raw-format gibberish" they'll get YUV? I'd have thought we really want to reject that instead.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -806,6 +844,40 @@ static int parse_cmdline(int argc, const char **argv, RASPIVID_STATE *state)
...
+
+         if (!state->raw_output)
+         {
+            fprintf(stderr, "--raw-fmt is only valid with --raw output selected\n");
+            valid = 0;
```
6by9: Ah the old chicken and egg problem.
```raspivid --raw filename.rgb --raw-format rgb``` is valid, but
```raspivid --raw-format rgb --raw filename.rgb ``` isn't because output wasn't set when they tried setting the format. There's never a good way around that, except possibly checking the two settings after they've all been parsed and then identifying the conflict.

And line 850 would mean that you'd get YUV with the second line anyway.

File: host_applications/linux/apps/raspicam/RaspiVid.c [hunk 4]
```diff
@@ -1289,6 +1367,74 @@ static void encoder_buffer_callback(MMAL_PORT_T *port, MMAL_BUFFER_HEADER_T *buf
...
+
+   // All our times based on the receipt of the first callback
+   if (base_time == -1)
+      base_time = vcos_getmicrosecs64()/1000;
+
+   // We pass our file handle and other stuff in via the userdata field.
+
+   PORT_USERDATA *pData = (PORT_USERDATA *)port->userdata;
+
+   if (pData)
```
6by9: Not a complaint about this review, but this lot really should get refactored as we now have it present about 4 times in the different callbacks.

host_applications/linux/apps/raspicam/RaspiVid.c
adamheinrich: I wanted to adapt to the previous code. Line 650 does it in a similar way. Ok, I will change this.

host_applications/linux/apps/raspicam/RaspiVid.c
adamheinrich: I will remove the redundant code (`base_time` is not used in this callback at all).

adamheinrich:
So I have changed the original commit again, referenced commits are from separate branch:

- Redundant code from `splitter_buffer_calback()` is removed ([commit](https://github.com/adamheinrich/userland/commit/ed3d7bfe69c1316da959358f3a105cdd19d18bf1))
- Wrong output format produces an error, `--raw-format` without `--raw` is ignored ([commit](https://github.com/adamheinrich/userland/commit/f03fd9a9e0826402a76e4973876507bee32d78fe))
- Help messages is modified ([commit](https://github.com/adamheinrich/userland/commit/325b341e131d8a74b2e1ccf6f975621a440afede))

adamheinrich:
I've just changed the original commit again (fixing typo in a comment: [commit](https://github.com/adamheinrich/userland/commit/ddd50008a897203ba757ff70a96ed8380bb326ed)).

6by9:
My sincerest apologies for not reviewing this sooner. Doing so now.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2103,6 +2392,14 @@ int main(int argc, const char **argv)
       destroy_camera_component(&state);
       exit_code = EX_SOFTWARE;
    }
+   else if (state.raw_output && (status = create_splitter_component(&state)) != MMAL_SUCCESS)
```
6by9: Build warning off this when cross-compiling:
userland/host_applications/linux/apps/raspicam/RaspiVid.c: In function ‘main’:
userland/host_applications/linux/apps/raspicam/RaspiVid.c:2412:30: warning: ‘status’ may be used uninitialized in this function [-Wmaybe-uninitialized]
    else if (state.raw_output && (status = create_splitter_component(&state)) != MMAL_SUCCESS)
The error isn't this line but the fact that create_splitter_component can return an uninitialised value - see above.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1558,6 +1691,159 @@ static void destroy_camera_component(RASPIVID_STATE *state)
...
+ *
+ * @return MMAL_SUCCESS if all OK, something else otherwise
+ *
+ */
+static MMAL_STATUS_T create_splitter_component(RASPIVID_STATE *state)
+{
+   MMAL_COMPONENT_T *splitter = 0;
+   MMAL_PORT_T *splitter_output = NULL;
+   MMAL_ES_FORMAT_T *format;
+   MMAL_STATUS_T status;
```
6by9: MMAL_STATUS_T status = MMAL_ENOSYS;
Otherwise we can return an uninitialised value from the if (state->camera_component == NULL) error path.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -141,10 +149,19 @@ typedef struct
...
    FILE *pts_file_handle;               /// File timestamps
 } PORT_USERDATA;
 
+/** Possible raw output formats
+ */
+typedef enum {
+   RAW_OUTPUT_FMT_YUV = 1,
+   RAW_OUTPUT_FMT_RGB,
+   RAW_OUTPUT_FMT_GRAY,
+} raw_output_fmt_t;
```
6by9: For consistency with the rest of this codebase, typedef'ed types should be capitals, so RAW_OUTPUT_FMT_T.

adamheinrich: Ok, will probably name it RAW_OUTPUT_FMT as other typedefs don't end with _T (PORT_USERDATA, RASPIVID_STATE).

adamheinrich:
So I've changed the original commit again, see individual commits ([1](https://github.com/adamheinrich/userland/commit/616141ddf0f0ec6c009921b14232a18614ac0358), [2](https://github.com/adamheinrich/userland/commit/5c2a7b1efeda02198f490663d039f11bea427ca4)) if you want.

6by9:
> By the way I get warning: ‘status’ may be used uninitialized in this function on line 1286 in the original file (and the same code in my new splitter_buffer_callback()).
But because you didn't report this warning (you probably have more clever compiler :) ) and the buildme script doesn't report it either I'll leave it untouched.

That one is actually safe:
```
if(new_buffer)
  status = ...
if(!new_buffer || status != MMAL_SUCCESS)
  vcos_log_error
```
So:
- if new_buffer then status will be valid.
- if !new_buffer, then the if statement will abort after checking that term and therefore not look at status. It therefore doesn't matter that it isn't initialised.
I did have to look at it several times to work out why the compiler wasn't objecting, but I think it is correct in not raising an error.

6by9:
I'm happy with all of these changes now, although I still haven't run them yet (cherrypicked and built, just not run).

PR #344, opened 2016-10-07
EGL VCSM buffers

timgover:
Enable zero-copy access to EGL textures and framebuffers from the CPU by adding an EGL extension that allocates buffers from VCSM.
Buffers must be RGBA raster order.
For normal 3D rendering it's much slower to use RSO buffers, however, for image processing it can be more efficient to use an RSO buffer if the buffer is only used once e.g. as part of an OpenCV pipeline.
This requires a beta version of the firmware.


6by9:
Can't really comment on the test app, but the other changes look OK.
Firmware changes need to be merged before these though.

Ruffio:
@6by9 Can this be merged now or we waiting for other firmware merges?

popcornmix:
@Ruffio Currently the CMakefile change can break existing makefiles (easy enough to fix up, but will potential break a number of applications until that is done).

@timgover is aware and is thinking of a solution.

timgover:
My plan is to change the dependency between the EGL library  and VCSM to be a pure run-time version i.e. dlopen. That way it shouldn't matter whether VCSM is present or not and should be resilient to random combinations of EGL libraries.

I think it's a pretty simple change because there are only a couple of function calls.

timgover:
I've removed the static library dependency between EGL and VCSM. It now loads libvcsm.so via dlopen so this should be compatible with Pi distributions without VCSM.

If anyone wants to have a static build dependency then modifying CMakeLists.txt to defined KHRONOS_HAVE_VCSM still works.

I verified this by building and verifying libraries with ldd and removing libvcsm.so

If there's a specific OS I should test with then please let me know, it's only verified on Raspbian Pixel so far.

timgover:
Minor tidyup to error handling in egl_init_vcsm. I think that case would only be possible if the application is statically linked with libvcsm.a but libvcsm.so is missing which seems unlikely.

popcornmix:
Thanks - seems to work for me and isn't breaking other libs using egl. Thanks!

PR #347, opened 2016-10-19
Add ARM64 CMake variable for 64-bit targets

rfinnie:
At the moment, much of the codebase compiles on 64-bit targets, but not
all (mostly 32-bit assembly).  -DARM64=ON allows for skipping
directories which currently do not compile correctly.

This is backwards compatible with existing behavior (i.e. no -DARM on a
32-bit target compiles everything as before), and hopefully will
ultimately become unnecessary as remaining components are fixed for
64-bit use.

FYI, everything not skipped by -DARM64=ON _compiles_ correctly (albeit with some warnings), but I haven't tested everything for actually working.  However I have verified vcgencmd and tvservice work fine with 64-bit kernel/userland.


ED6E0F17:
There is existing C code to replace the ARMv6 assemly at "interface/khronos/common/khrn_int_hash.c", but I don`t know enough about CMAKE to conditionally disable https://github.com/raspberrypi/userland/blob/master/interface/khronos/CMakeLists.txt#L55

Electron752:
Actually, I was able to get this to work just fine.  So maybe the PR should be accepted.

Here are the build steps: I didn't use buildme
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=release -DARM64=ON ../
make -j8

Rather cool actually!

Electron752:
I'm even able to run some of the commands:
./tvservice -s
state 0x120009 [HDMI CEA (16) RGB lim 16:9], 1920x1080 @ 60.00Hz, progressive
./tvservice -a
    PCM supported: Max channels: 2, Max samplerate:  32kHz, Max samplesize 16 bits.

Here is the list of binary that actually build:
-rwxr-xr-x 1 root root 12400 Oct 29 21:32 dtmerge
-rwxr-xr-x 1 root root 33584 Oct 29 21:32 dtoverlay
-rwxr-xr-x 1 root root 60328 Oct 29 21:33 tvservice
-rwxr-xr-x 1 root root 52200 Oct 29 21:33 vcfiled
-rwxr-xr-x 1 root root 13328 Oct 29 21:32 vcfiled_lock_test
-rwxr-xr-x 1 root root 20920 Oct 29 21:33 vcgencmd
-rwxr-xr-x 1 root root 67320 Oct 29 21:33 vchiq_test
-rwxr-xr-x 1 root root 10456 Oct 29 21:33 vcmailbox

rfinnie:
Yeah, a good chunk of the userland codebase works fine on ARM64, when used with @Electron752 's 64-bit kernel base.  This PR simply adds the ability to do -DARM64=ON to skip the code which is currently broken.

There's a certainly a lot to fix and some might be simple for people with better CMAKE experience (e.g. @ED6E0F17 pointing out that khronos might be workable if it uses the .c version of a file instead of the ASM version), but this should not hold up accepting this PR.  Ideally this PR is accepted first to at least allow for 64-bit compilation, and as things are fixed by people in subsequent PRs, what -DARM64=ON masks out is slowly reduced.

FYI, this patch is currently in use to enable arm64 .deb compilation at https://launchpad.net/~ubuntu-raspi2/+archive/ubuntu/ppa-nightly/+packages

PR #349, opened 2016-10-27
hello_pi: Add hello_mmal_encode example

sconemad:
An example showing how to use the image encoder via MMAL to encode a test image into various formats.

### DIFF
--- a/host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
+++ b/host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
@@ -1,0 +1,261 @@
+/*
+Copyright (C) 2016 RealVNC Limited. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+this list of conditions and the following disclaimer in the documentation
+and/or other materials provided with the distribution.
+
+3. Neither the name of the copyright holder nor the names of its contributors
+may be used to endorse or promote products derived from this software without
+specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+// Image encoding example using MMAL
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include <bcm_host.h>
+#include <interface/mmal/mmal.h>
+#include <interface/mmal/mmal_format.h>
+#include <interface/mmal/util/mmal_default_components.h>
+#include <interface/mmal/util/mmal_component_wrapper.h>
+#include <interface/mmal/util/mmal_util_params.h>
+
+// Format for test image
+const unsigned int WIDTH = 512;
+const unsigned int HEIGHT = 512;
+const MMAL_FOURCC_T INPUT_ENC = MMAL_ENCODING_RGBA;
+const unsigned int BYTESPP = 4;
+const uint32_t RED = 0xff;
+const uint32_t GREEN = 0xff << 8;
+const uint32_t BLUE = 0xff << 16;
+const uint32_t ALPHA = 0xff << 24;
+
+static MMAL_WRAPPER_T* encoder;
+static VCOS_SEMAPHORE_T sem;
+
+// This callback and the above semaphore is used when waiting for
+// data to be returned.
+static void mmalCallback(MMAL_WRAPPER_T* encoder)
+{
+   vcos_semaphore_post(&sem);
+}
+
+// Create a test input image in the supplied buffer consisting of 8 vertical bars of
+// white | black | red | green | blue | cyan | magenta | yellow
+void create_rgba_test_image(void* buf, unsigned int length, unsigned int stride)
+{
+   uint32_t* pixel = buf;
+   int i;
+   for (i=0; i<length/BYTESPP; ++i) {
+      switch ((i % stride) / (WIDTH / 8)) {
+      case 0: *pixel = RED | GREEN | BLUE; break;
+      case 1: *pixel = 0; break;
+      case 2: *pixel = RED; break;
+      case 3: *pixel = GREEN; break;
+      case 4: *pixel = BLUE; break;
+      case 5: *pixel = GREEN | BLUE; break;
+      case 6: *pixel = RED | BLUE; break;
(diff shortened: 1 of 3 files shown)


host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+  MMAL_STATUS_T status;
+  int eos = 0;
+  int sent = 0;
+  int outputWritten = 0;
+  int outFile;
+  int i, nw;
+  
+  fprintf(stdout, "Encoding test image %s\n", filename);
+
+  if (mmal_wrapper_create(&mmal, MMAL_COMPONENT_DEFAULT_IMAGE_ENCODER)
```
6by9: mmal as a variable name is very ambiguous. "encoder" or "component"?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+    exit(1);
+  }
+  mmal->callback = mmalCallback;
+
+  // Configure input
+
+  portIn = mmal->input[0];
+  mmal->status = MMAL_SUCCESS;
+
+  if (portIn->is_enabled) {
```
6by9: I'm not aware of any way for components to have enabled ports this soon after creation. Redundant? Or is this copy/paste from other examples? (In which case they ought to be amended).

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+  
+  fprintf(stdout, "- input %4.4s %ux%u\n",
+          (char*)&portIn->format->encoding,
+          portIn->format->es->video.width, portIn->format->es->video.height);
+
+  // Configure output
+  
+  portOut = mmal->output[0];
+
+  if (portOut->is_enabled) {
```
6by9: Ditto - I don't see how this can happen.

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+  if (outFile < 0) {
+    fprintf(stderr, "Failed to open file: %s\n", filename);
+    exit(1);
+  }
+  
+  while (!eos) {
+
+    // Send output buffers to be filled with encoded image.
+    while (mmal_wrapper_buffer_get_empty(portOut, &out, 0) == MMAL_SUCCESS) {
+      out->data = outBuf;
```
6by9: This sounds like it is going to be very bad if portOut->buffer_num > 1 as you'll use the same pointer for all the buffers.

popcornmix:
I suggested the name for this, so I'm okay with it.
If one day we have a more complete set of openmax/mmal image/video encode/decode examples we can come up with a new naming scheme (perhaps partly directory and partly filename) and this can be renamed.

6by9:
> I suggested the name for this, so I'm okay with it.

:-)
I'll add the porting of the IL examples to MMAL to my list of jobs - it's the sort of trivial task that can be done without too much concentration of an evening, unlike some of the others. When done I'll discuss with you over directory/naming.

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+    fprintf(stderr, "Failed to enable output port\n");
+    exit(1);
+  }
+
+  fprintf(stdout, "- output %4.4s\n", (char*)&encoding);
+  
+  // Perform the encoding
+
+  outBuf = malloc(portOut->buffer_size);
+  outFile = open(filename, O_WRONLY | O_CREAT | O_TRUNC, 0644);
```
luked99: Is there a reason for using open() rather than fopen() ? Won't open() mean that the output is unbuffered?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+  if (outFile < 0) {
+    fprintf(stderr, "Failed to open file: %s\n", filename);
```
luked99: Would be nice to print strerror(errno) here as well, to help people work out why it didn't succeed (same comment elsewhere).

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+      out->alloc_size = portOut->buffer_size;
+      if (mmal_port_send_buffer(portOut, out) != MMAL_SUCCESS) {
+        fprintf(stderr, "Failed to send buffer\n");
+        break;
+      }
+    }
+
+    // Send image to be encoded.
+    if (!sent && mmal_wrapper_buffer_get_empty(portIn, &in, 0) == MMAL_SUCCESS) {
+      fprintf(stdout, "- sending %u bytes to encoder\n", inBufSize);
```
luked99: Perhaps just printf() rather than fprintf(stdout,...) ?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+      memcpy(in->data, inBuf, inBufSize);
```
luked99: Is this ending up copying all of the input stream? Is there any way to avoid that?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+unsigned int WIDTH = 512;
+unsigned int HEIGHT = 512;
+MMAL_FOURCC_T INPUT_ENC = MMAL_ENCODING_RGBA;
+unsigned int BYTESPP = 4;
+uint32_t RED = 0xff;
+uint32_t GREEN = 0xff << 8;
+uint32_t BLUE = 0xff << 16;
+uint32_t ALPHA = 0xff << 24;
+
+VCOS_SEMAPHORE_T sem;
```
luked99: Could this be static?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
```diff
@@ -0,0 +1,268 @@
...
+  vcos_semaphore_post(&sem);
+}
+
+// mmal_encode_test - Encode a test image and write to file
+void mmal_encode_test(MMAL_FOURCC_T encoding, // Encoding
+                      const char* filename) // File name
+{
+  MMAL_WRAPPER_T* mmal;
+  unsigned int inBufSize;
+  unsigned char* inBuf;
```
luked99: If you made inBuf a void* then you wouldn't need the ugly casting later on.

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
sconemad: Yes, it doesn't seem to be necessary in this case. In general it might be though, if people are reconfiguring an existing encoder, so perhaps it would be good to show this. How about if the encoder is created in main(), then reconfigured for each encoding?

host_applications/linux/apps/hello_pi/hello_mmal_encode/mmal_encode.c
sconemad: I could construct the test image directly in the in->data buffer.
I'm also assuming the size of this buffer is always going to be >= inBuffSize (which it does always seem to be), but not sure if this is guaranteed?

sconemad:
Hi, thanks for all the comments. I've pushed a further commit which hopefully addresses these.

popcornmix:
@6by9 @luked99 okay with this now?

luked99:
Would you mind squashing those two commits together into one? I think it makes it easier to review.

I haven't actually tried to run it yet, but it looks pretty convincing to me. But a few more comments (sorry!).

Is the indentation intended to be just two spaces? On other code in the same directory it's 3 spaces (and either 4, or a hard tab would be my choice, but that's a flame war too far :-) But consistent with the rest of the code perhaps?

Should this be renamed to something like hello_mmal_image_encode, in case anyone ever adds a video encode example?

6by9:
@luked99 
> Is the indentation intended to be just two spaces? In other code in the same directory it's 3 spaces (and either 4, or a hard tab would be my choice, but that's a flame war too far :-) But consistent with the rest of the code perhaps?

Consistent would be good and appears to have been done. As you know, 3 was the Broadcom standard hence why it was applied to these files.
Hard tabs: you've been doing too much kernel hacking ;-) You'll be limiting line length to 80 chars next....

> Should this be renamed to something like hello_mmal_image_encode, in case anyone ever adds a video encode example?

I already asked that question of popcornmix https://github.com/raspberrypi/userland/pull/349#pullrequestreview-6082438 - he'd suggested the name :-)
I will start looking at porting the IL examples to MMAL, at which point we can put this and others in a MMAL directory and come up with a consistent naming scheme.

popcornmix:
Many thanks @sconemad
Hopefully this will be the start of a collection of mmal examples that will help other users.

PR #350, opened 2016-10-28
Fix the minimum amount of pointer size related issues in vchiq_test

Electron752:
vchiq_test is critical for testing vc04_services in the drivers/staging branch of the upstream kernel.   That version is going to support arm64, so it's necessary to fix the critical pointer related issues in this test.

For now, I am not attempting to make public modifications of the build scripts or fix all the pointer related warnings.   Just fix the critical issues to get the test to pass.

### DIFF
--- a/interface/vchiq_arm/vchiq_test.c
+++ b/interface/vchiq_arm/vchiq_test.c
@@ -1201,15 +1201,15 @@ func_data_test(VCHIQ_SERVICE_HANDLE_T service, int datalen, int align, int serve
    EXPECT(vchiq_queue_message(service, &element, 1), VCHIQ_SUCCESS);
 
    memset(databuf, 0xff, sizeof(databuf));
-   data = (uint8_t *)((uint32_t)databuf & ~(PAGE_SIZE - 1)) + align;
-   data = (uint8_t *)((((int)databuf + PROLOGUE_SIZE) & ~(FUN2_MAX_ALIGN - 1)) + align - PROLOGUE_SIZE);
+   data = (uint8_t *)((uintptr_t)databuf & ~(PAGE_SIZE - 1)) + align;
+   data = (uint8_t *)((((intptr_t)databuf + PROLOGUE_SIZE) & ~(FUN2_MAX_ALIGN - 1)) + align - PROLOGUE_SIZE);
    if (data < databuf)
       data += PAGE_SIZE;
    data += PROLOGUE_SIZE;
 
    EXPECT(vchiq_queue_bulk_receive(service, data, datalen, NULL), VCHIQ_SUCCESS);
 
-   data2 = (uint8_t *)(((uint32_t)databuf2 + PROLOGUE_SIZE) & ~(PAGE_SIZE - 1)) + align - PROLOGUE_SIZE;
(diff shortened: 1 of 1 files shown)


interface/vchiq_arm/vchiq_test.c
```diff
@@ -1201,15 +1201,15 @@ func_data_test(VCHIQ_SERVICE_HANDLE_T service, int datalen, int align, int serve
    EXPECT(vchiq_queue_message(service, &element, 1), VCHIQ_SUCCESS);
 
    memset(databuf, 0xff, sizeof(databuf));
-   data = (uint8_t *)((uint32_t)databuf & ~(PAGE_SIZE - 1)) + align;
-   data = (uint8_t *)((((int)databuf + PROLOGUE_SIZE) & ~(FUN2_MAX_ALIGN - 1)) + align - PROLOGUE_SIZE);
+   data = (uint8_t *)((unsigned long)databuf & ~(PAGE_SIZE - 1)) + align;
+   data = (uint8_t *)((((long)databuf + PROLOGUE_SIZE) & ~(FUN2_MAX_ALIGN - 1)) + align - PROLOGUE_SIZE);
```
luked99: Should we be using intptr_t and uintptr_t these days?

http://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdint.h.html

interface/vchiq_arm/vchiq_test.c
```diff
@@ -1319,6 +1319,8 @@ clnt_callback(VCHIQ_REASON_T reason, VCHIQ_HEADER_T *header,
    vcos_mutex_lock(&g_mutex);
    if (reason == VCHIQ_MESSAGE_AVAILABLE)
    {
+      unsigned long header_size = header->size;
```
luked99: size_t here?

interface/vchiq_arm/vchiq_test.c
```diff
@@ -1633,7 +1635,7 @@ static void check_timer(void)
 
 static char *buf_align(char *buf, int align_size, int align)
 {
-   char *aligned = buf - ((int)buf & (align_size - 1)) + align;
+   char *aligned = buf - ((long)buf & (align_size - 1)) + align;
```
luked99: inptr_t again?

Electron752:
OK, I changed the casts to uintptr_t/intptr_t.

See the updated pull request.

Electron752:
I probably should have put a better description of the (header->size != 0) issue.  It's actually a race condition.

In the few lines before the message is released.  That clears out the message which is causing the following test to fail even though it was a valid message.

I think the test could use a bunch of work in general.  I'm just trying to band-aid it into getting it to work.  Not fix all the issues.

Electron752:
I guess I could do a sweep and fix all the warnings, but I'm not sure it's best to stop progress just because the progress doesn't fix all the issues.

Electron752:
If some "high-level" decision needs to be made on this, I'm more then happy to close out the pull request until I fix all the issues.

This tool doesn't appear to be changing much, so another possibility is for me to just keep it on a separate github.com repo and just share it with the people that are interested at this time.

luked99:
Maybe just split this change into two separate parts so that it's clear what's going on. I think making a high level decision is way above my paygrade, but I think it would be OK to just submit your changes piecemeal as they develop. Changing some casts from (long/int) to (uintptr_t) isn't too controversial.

PR #351, opened 2016-10-29
Add end bracket for cpp extern

peter-harliman:
The missing bracket would cause compilation error in cpp.


luked99:
Looks good, thanks!

PR #352, opened 2016-10-29
buildme: fix whitespace

luked99:
No changes other than whitespace.

Thinking about it, I think buildme needs a bit of attention: it should presumably have a way to build for arm64; why only 6 jobs.

### DIFF
--- a/buildme
+++ b/buildme
@@ -2,8 +2,8 @@
 BUILDTYPE=Release
 
 if [ "$1" = "--debug" ]; then
-BUILDTYPE=Debug
-shift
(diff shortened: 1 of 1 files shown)


Electron752:
I think all that needs to be done for arm64 is disable building components that are asm based.   Those components don't make sense and as far as I know, no plans are in place to ever support a 64 bit version of Raspbian.

That said some of the tools are still useful even on other distributions.

I would submit a pull request to make the changes, just I don't know that much about cmake to make the changes in a way that other will agree upon.

PR #353, opened 2016-10-30
Fix arm64 issues vchiq_test

Electron752:
vchiq_test is critical for testing vchiq on the upstream kernel which is currently in staging and will support arm64.   This pull request fixes the minimum number of issues to get vchiq_test to run on arm64.

1. Fixes a few casts to use uintptr_t and intptr_t.
2. Fix a race condition which seems to happen the most on arm64 for some reason.

NOTE: I am not attempting to fix the build scripts at this time.

### DIFF
--- a/interface/vchiq_arm/vchiq_test.c
+++ b/interface/vchiq_arm/vchiq_test.c
@@ -1201,15 +1201,15 @@ func_data_test(VCHIQ_SERVICE_HANDLE_T service, int datalen, int align, int serve
    EXPECT(vchiq_queue_message(service, &element, 1), VCHIQ_SUCCESS);
 
    memset(databuf, 0xff, sizeof(databuf));
-   data = (uint8_t *)((uint32_t)databuf & ~(PAGE_SIZE - 1)) + align;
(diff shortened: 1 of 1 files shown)


luked99:
Looks good to me.

popcornmix:
Looks okay from here. @pelwell ?

Electron752:
Any issues here?  The changes should be small and self contained.

NOTE: vchiq on arm64 is planned to still work from a 32 bit arm userland app(such as 32 bit vchiq ).  The ioctl data munging is still not completed so that the only way to fully test on arm64 at this point is vchiq_test and some of the other tools in userland such as tvservice.

luked99:
> Any issues here? The changes should be small and self contained.

They look good to me, the buf->size problem looks obviously an improvement.

I think it's possibly just worth giving others a chance to comment before merging.

Electron752:
OK, sounds good.

I need to get away from the PC for awhile.  Let me know if any issues come up.   

I don't think anything I've submitted should be that complex/controversial.  Like I said, the plan is to eventually get 32 bit vchiq_test to work on an arm64 kernel as well as 64 bit vchiq_test.  It's just a matter of munging through the code on the vchiq driver.

pelwell:
I'm happy with both changes - thanks.

PR #354, opened 2016-11-11
3 minor raspicam updates

6by9:
Setting the restart interval isn't supported on older firmwares, so aborting is a touch excessive as long as it isn't set.

Segmentation option could try opening a null filename.

Option to segment timestamps file as well as encoded data and motion vectors.


JamesH65:
Looks good to me.

PR #357, opened 2016-11-16
Fix enum conversion warnings

kraj:


6by9:
Looks reasonable to me, though I'm not aware of anything really using this library.

PR #358, opened 2016-11-19
vcfilesys.c: Use PRId64 c99 type for printing 64bit integer in portab…

kraj:
…le way

Fixes building this file on 64bit arm
error: format '%lld' expects argument of type 'long long int', but argument 4 has type 'int64_t {aka long int}'

### DIFF
--- a/interface/vmcs_host/linux/vcfilesys.c
+++ b/interface/vmcs_host/linux/vcfilesys.c
@@ -243,7 +243,7 @@ RETURNS
 
 int64_t vc_hostfs_lseek64(int fildes, int64_t offset, int whence)
 {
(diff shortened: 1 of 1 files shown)


6by9:
Looks good to me.

More general question: Is there actually any point in building vcfiled for Pi? I don't believe anything on the VPU uses it, and I don't know if the released builds include host FS in the firmware. Might be a candidate for removal.

popcornmix:
Yes, vcfiled hasn't been used for years (it was used for loading vlls when the Pi1 first came out).

popcornmix:
No objections to this for now (although the file may disappear in the future).

PR #359, opened 2016-12-02
RaspiVid is now capable to listen on a certain TCP port and wait for …

maroviher:
…an incoming connection. The same functionality as

raspivid -o - | nc -k -l -p 5001
but without nc invokation and without pipe read/write overhead. It reduces cpu usage and video latency. Added help message for an output to a network.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -52,7 +52,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 // We use some GNU extensions (basename)
-#define _GNU_SOURCE
+#ifndef _GNU_SOURCE
+   #define _GNU_SOURCE
+#endif
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -83,6 +85,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "RaspiCLI.h"
 
 #include <semaphore.h>
+
+#include <stdbool.h>
 
 // Standard port setting for the camera component
 #define MMAL_CAMERA_PREVIEW_PORT 0
@@ -226,6 +230,8 @@ struct RASPIVID_STATE_S
    int save_pts;
    int64_t starttime;
    int64_t lasttime;
+
+   bool netListen;
 };
 
 
@@ -316,6 +322,7 @@ static void display_valid_parameters(char *app_name);
 #define CommandLevel        31
 #define CommandRaw          32
 #define CommandRawFormat    33
+#define CommandNetListen    34
 
 static COMMAND_LIST cmdline_commands[] =
 {
@@ -323,7 +330,10 @@ static COMMAND_LIST cmdline_commands[] =
    { CommandWidth,         "-width",      "w",  "Set image width <size>. Default 1920", 1 },
    { CommandHeight,        "-height",     "h",  "Set image height <size>. Default 1080", 1 },
    { CommandBitrate,       "-bitrate",    "b",  "Set bitrate. Use bits per second (e.g. 10MBits/s would be -b 10000000)", 1 },
-   { CommandOutput,        "-output",     "o",  "Output filename <filename> (to write to stdout, use '-o -')", 1 },
+   { CommandOutput,        "-output",     "o",  "Output filename <filename> (to write to stdout, use '-o -').\n"
+         "\t\t  Connect to a remote IPv4 host (e.g. tcp://192.168.1.2:1234, udp://192.168.1.2:1234)\n"
+         "\t\t  To listen on a TCP port (IPv4) and wait for an incoming connection use -l\n"
+         "\t\t  (e.g. raspvid -l -o tcp://0.0.0.0:3333 -> bind to all network interfaces, raspvid -l -o tcp://192.168.1.1:3333 -> bind to a certain local IPv4)", 1 },
    { CommandVerbose,       "-verbose",    "v",  "Output verbose information during run", 0 },
    { CommandTimeout,       "-timeout",    "t",  "Time (in ms) to capture for. If not specified, set to 5s. Zero to disable", 1 },
    { CommandDemoMode,      "-demo",       "d",  "Run a demo mode (cycle through range of camera options, no capture)", 1},
@@ -353,6 +363,7 @@ static COMMAND_LIST cmdline_commands[] =
    { CommandLevel,         "-level",      "lev","Specify H264 level to use for encoding", 1},
    { CommandRaw,           "-raw",        "r",  "Output filename <filename> for raw video", 1 },
    { CommandRawFormat,     "-raw-format", "rf", "Specify output format for raw video. Default is yuv", 1},
+   { CommandNetListen,     "-listen",     "l", "Listen on a TCP socket", 0},
 };
 
 static int cmdline_commands_size = sizeof(cmdline_commands) / sizeof(cmdline_commands[0]);
@@ -427,6 +438,8 @@ static void default_status(RASPIVID_STATE *state)
 
    state->frame = 0;
    state->save_pts = 0;
+
+   state->netListen = false;
 
 
    // Setup preview window defaults
@@ -873,6 +886,13 @@ static int parse_cmdline(int argc, const char **argv, RASPIVID_STATE *state)
          break;
       }
 
+      case CommandNetListen:
+      {
+         state->netListen = true;
+
+         break;
+      }
+
       default:
       {
          // Try parsing for any image specific parameters
@@ -1034,42 +1054,115 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
 
    if (filename)
    {
-      int network = 0, socktype;
+      bool bNetwork = false;
+      int sfd, socktype;
+
       if(!strncmp("tcp://", filename, 6))
       {
-         network = 1;
+         bNetwork = true;
          socktype = SOCK_STREAM;
       }
-      if(!strncmp("udp://", filename, 6))
+      else if(!strncmp("udp://", filename, 6))
       {
-         network = 1;
+         if (pState->netListen)
+         {
+            fprintf(stderr, "No support for listening in UDP mode\n");
+            exit(131);
+         }
+         bNetwork = true;
          socktype = SOCK_DGRAM;
       }
-      if(network)
+
+      if(bNetwork)
       {
+         unsigned short port;
          filename += 6;
          char *colon;
-         colon = strchr(filename, ':');
-         int sfd = socket(AF_INET, socktype, 0);
-         if(sfd < 0)
+         if(NULL == (colon = strchr(filename, ':')))
          {
-              perror("socket");
+            fprintf(stderr, "%s is not a valid IPv4:port, use something like tcp://1.2.3.4:1234 or udp://1.2.3.4:1234\n",
+                    filename);
+            exit(132);
          }
-
-         unsigned short port;
-         sscanf(colon + 1, "%hu", &port);
+         if(1 != sscanf(colon + 1, "%hu", &port))
+         {
+            fprintf(stderr,
+                    "Port parse failed. %s is not a valid network file name, use something like tcp://1.2.3.4:1234 or udp://1.2.3.4:1234\n",
+                    filename);
+            exit(133);
+         }
+         char chTmp = *colon;
          *colon = 0;
 
-         fprintf(stderr, "Connecting to %s:%hu\n", filename, port);
-         
-         struct sockaddr_in saddr;
+         struct sockaddr_in saddr={};
          saddr.sin_family = AF_INET;
          saddr.sin_port = htons(port);
-         inet_aton(filename, &saddr.sin_addr);
-
-         if(connect(sfd, (struct sockaddr *)&saddr, sizeof(struct sockaddr_in)) < 0)
+         if(0 == inet_aton(filename, &saddr.sin_addr))
          {
-            perror("connect");
+            fprintf(stderr, "inet_aton failed. %s is not a valid IPv4 address\n",
+                    filename);
+            exit(134);
+         }
+         *colon = chTmp;
+
+         if (pState->netListen)
+         {
+            int sockListen = socket(AF_INET, SOCK_STREAM, 0);
+            if (sockListen >= 0)
+            {
+               int iTmp = 1;
+               setsockopt(sockListen, SOL_SOCKET, SO_REUSEADDR, &iTmp, sizeof(int));//no error handling, just go on
+               if (bind(sockListen, (struct sockaddr *) &saddr, sizeof(saddr)) >= 0)
+               {
+                  while ((-1 == (iTmp = listen(sockListen, 0))) && (EINTR == errno))
+                     ;
+                  if (-1 != iTmp)
+                  {
+                     fprintf(stderr, "Waiting for a TCP connection on %s:%"SCNu16"...",
+                             inet_ntoa(saddr.sin_addr), ntohs(saddr.sin_port));
+                     struct sockaddr_in cli_addr;
+                     socklen_t clilen = sizeof(cli_addr);
+                     while ((-1 == (sfd = accept(sockListen, (struct sockaddr *) &cli_addr, &clilen))) && (EINTR == errno))
+                        ;
+                     if (sfd >= 0)
+                        fprintf(stderr, "Client connected from %s:%"SCNu16"\n", inet_ntoa(cli_addr.sin_addr), ntohs(cli_addr.sin_port));
+                     else
+                        fprintf(stderr, "Error on accept: %s\n", strerror(errno));
+                  }
+                  else//if (-1 != iTmp)
+                  {
+                     fprintf(stderr, "Error trying to listen on a socket: %s\n", strerror(errno));
+                  }
+               }
+               else//if (bind(sockListen, (struct sockaddr *) &saddr, sizeof(saddr)) >= 0)
+               {
+                  fprintf(stderr, "Error on binding socket: %s\n", strerror(errno));
+               }
+            }
+            else//if (sockListen >= 0)
+            {
+               fprintf(stderr, "Error creating socket: %s\n", strerror(errno));
+            }
+
+            if (sockListen >= 0)//regardless success or error
+               close(sockListen);//do not listen on a given port anymore
+         }
+         else//if (pState->netListen)
+         {
+            if(0 <= (sfd = socket(AF_INET, socktype, 0)))
+            {
+               fprintf(stderr, "Connecting to %s:%hu...", inet_ntoa(saddr.sin_addr), port);
+
+               int iTmp = 1;
+               while ((-1 == (iTmp = connect(sfd, (struct sockaddr *) &saddr, sizeof(struct sockaddr_in)))) && (EINTR == errno))
+                  ;
+               if (iTmp < 0)
+                  fprintf(stderr, "error: %s\n", strerror(errno));
+               else
+                  fprintf(stderr, "connected, sending video...\n");
+            }
+            else
+               fprintf(stderr, "Error creating socket: %s\n", strerror(errno));
          }
 
          new_handle = fdopen(sfd, "w");


JamesH65:
Not tested, but looks sane. Not a network sockets expert though, so cannot
comment further.

Ferroin:
I've just taken a cursory look myself (I don't have the hardware to test this), but everything looks correct and secure.

One general comment though, instead of adding 'l' to the beginning of stuff to specify listening, I would suggest using a separate switch, as that's more consistent with both netcat, and quite a few other tools.

maroviher:
@Ferroin:

I would distinguish between '-o' (for write to a regular file or stdout) and some new switch for writing to a network, but we have already tcp:// and udp:// prefixes for '-o', so I thought a tcp:// could be a minimal change to a already existing switch.

Ferroin:
My main complaint is that it's not intuitive (ltcp:// looks like a different protocol from tcp://, but it isn't), and it's inconsistent with most other tools.  Take a look at nc, or iperf, or other tools that can listen or connect to a remote listener.  Essentially all of them use a switch to determine whether to run as a server or client, because it makes things clearer, and it's a whole lot safer to parse a switch than to parse a free-form argument.

JamesH65:
Out of interest, what does this change actually do over and above the tcp
option?

maroviher:
@JamesH65 
This change (ltcp) would wait for an incoming TCP connection infinitely and only after a client is connected start the camera and send a data over a network. This is useful if a client is behind a NAT.
The already existing (tcp) option would try to initiate a TCP connection to a given host immediately. In this case you do not have any chance to communicate with a client behind a NAT.

@Ferroin 
Got you finally.

popcornmix:
I think a separate command line option for listen mode is more consistent with how netcat and similar tools work, so that would be my preference.

JamesH65:
Inclined to agree - this is actually waiting on input, which isn't really
appropriate for a output switch.

maroviher:
@popcornmix 
How about using the '-o' option for regular files only and a new parmeter '-on' for a network streaming? We could use both parameters at the same time. And the third parameter '-l' to specify listening. 
raspivid -o myfile -on tcp://0.0.0.0:1234 -l
raspivid -o myfile -on tcp://192.168.1.2:1234

Howewer, It would not be backwards compatible to the current version, where '-o'  can be used with tcp:// and udp:// prefixes. Where one must say, you do not have any chance to learn about this prefixes in the current version, because nothing about that is in the usage text. I have come to it by studying of the source code.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -323,7 +323,9 @@ static COMMAND_LIST cmdline_commands[] =
    { CommandWidth,         "-width",      "w",  "Set image width <size>. Default 1920", 1 },
    { CommandHeight,        "-height",     "h",  "Set image height <size>. Default 1080", 1 },
    { CommandBitrate,       "-bitrate",    "b",  "Set bitrate. Use bits per second (e.g. 10MBits/s would be -b 10000000)", 1 },
-   { CommandOutput,        "-output",     "o",  "Output filename <filename> (to write to stdout, use '-o -')", 1 },
+   { CommandOutput,        "-output",     "o",  "Output filename <filename> (to write to stdout, use '-o -').\n"
+         "\t\t  Connect to a remote Host (e.g. tcp://192.168.1.2:1234, udp://192.168.1.2:1234)\n"
+         "\t\t  Listen on a TCP port and wait for an incoming connection (e.g. ltcp://0.0.0.0:1234, ltcp://192.168.1.1:1234)", 1 },
```
luked99: LTCP seems like an unfortunate choice, since it is an existing (albeit draft, and now forgotten) protocol:

https://tools.ietf.org/html/draft-bhandarkar-ltcp-00

A separate option to enable this might (as others have said) be better.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
-         int sfd = socket(AF_INET, socktype, 0);
-         if(sfd < 0)
+         const char ltcpSrch[] = "ltcp://";
+         if (!strncmp(ltcpSrch, filename, sizeof(ltcpSrch) - 1))
          {
-              perror("socket");
-         }
+            network = network_listen = 1;
+            struct sockaddr_in serv_addr, cli_addr;
+            bzero((char *) &serv_addr, sizeof(serv_addr));
```
luked99: memset is preferred over bzero:

http://pubs.opengroup.org/onlinepubs/009695399/functions/bzero.html

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+               goto _error;
+            }
 
-         unsigned short port;
-         sscanf(colon + 1, "%hu", &port);
-         *colon = 0;
+            int sockListen = socket(AF_INET, SOCK_STREAM, 0);
+            if (sockListen < 0)
+            {
+               fprintf(stderr, "error %d creating socket\n", errno);
```
luked99: It would be helpful to also convert errno to a string with strerror().

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+            {
+               close(sockListen);
+               fprintf(stderr, "error %d on listen\n", errno);
+               goto _error;
+            }
+            socklen_t clilen = sizeof(cli_addr);
+            sfd = accept(sockListen, (struct sockaddr *) &cli_addr, &clilen);
+            if (sfd < 0)
+            {
+               close(sockListen);
```
luked99: You can move close() to the error handling at the end.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+            int sockListen = socket(AF_INET, SOCK_STREAM, 0);
+            if (sockListen < 0)
+            {
+               fprintf(stderr, "error %d creating socket\n", errno);
+               goto _error;
+            }
+
+            serv_addr.sin_family = AF_INET;
+            serv_addr.sin_port = htons(serv_addr.sin_port);
+            int true = 1;
```
luked99: Using "true" as a variable name seems brave. It probably won't compile under C99.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+               close(sockListen);
+               fprintf(stderr, "error %d on listen\n", errno);
+               goto _error;
+            }
+            socklen_t clilen = sizeof(cli_addr);
+            sfd = accept(sockListen, (struct sockaddr *) &cli_addr, &clilen);
+            if (sfd < 0)
+            {
+               close(sockListen);
+               fprintf(stderr, "error %d on accept\n", errno);
```
luked99: Would be nicer to report strerror(errno).

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+         if (!network_listen)
          {
-            perror("connect");
+            filename += 6;
+            char *colon;
+            colon = strchr(filename, ':');
+            sfd = socket(AF_INET, socktype, 0);
+            if (sfd < 0)
+            {
+               perror("socket");
```
luked99: Error handling here?

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+            fprintf(stderr, "Connecting to %s:%hu\n", filename, port);
+
+            struct sockaddr_in saddr;
+            saddr.sin_family = AF_INET;
+            saddr.sin_port = htons(port);
+            inet_aton(filename, &saddr.sin_addr);
+
+            if (connect(sfd, (struct sockaddr *) &saddr, sizeof(struct sockaddr_in)) < 0)
+            {
+               perror("connect");
```
luked99: Error handling?

Strictly speaking this could also just return EINTR and require retrying on a signal. There's a rather ugly gcc macro to hide this.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+            }
 
-         fprintf(stderr, "Connecting to %s:%hu\n", filename, port);
-         
-         struct sockaddr_in saddr;
-         saddr.sin_family = AF_INET;
-         saddr.sin_port = htons(port);
-         inet_aton(filename, &saddr.sin_addr);
+            fprintf(stderr, "Client connected from %s:%"SCNu16"\n", inet_ntoa(cli_addr.sin_addr), ntohs(cli_addr.sin_port));
+            close(sockListen); //do not listen on a given port anymore
```
luked99: Could combine all error handling in one place.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
-         char *colon;
-         colon = strchr(filename, ':');
-         int sfd = socket(AF_INET, socktype, 0);
-         if(sfd < 0)
+         const char ltcpSrch[] = "ltcp://";
+         if (!strncmp(ltcpSrch, filename, sizeof(ltcpSrch) - 1))
          {
-              perror("socket");
-         }
+            network = network_listen = 1;
```
luked99: Could we use bool here instead? We're using other C99 features in this code.

host_applications/linux/apps/raspicam/RaspiVid.c
maroviher: I agree, see my previous post.

host_applications/linux/apps/raspicam/RaspiVid.c
maroviher: No, I can not, because at the end is not only an error handling.

host_applications/linux/apps/raspicam/RaspiVid.c
maroviher: This is not my, but an already signed-off code, on 24 august 2015, see git log.

popcornmix:
As we only support tcp for listening, and the IP address is unused, then only the port is valid.
Would it make sense to leave existing output options ('-o') the same, but add ``-l <port>`` as the new option. E.g. 
```raspivid -l 1234``` (listen for connection on port 1234)
```raspivid -o tcp://192.168.1.2:1234``` (send to 192.168.1.2:1234)

File: host_applications/linux/apps/raspicam/RaspiVid.c (hunk 8 above)
Comment (maroviher) [inline host_applications/linux/apps/raspicam/RaspiVid.c]: Yes, each system call can return EINTR. Should we add EINTR handling around each system call? What macro is that?
The same, highlighted code is already signed-off

File: host_applications/linux/apps/raspicam/RaspiVid.c [hunk 11]
```diff
@@ -1034,42 +1036,105 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+            struct sockaddr_in serv_addr, cli_addr;
+            bzero((char *) &serv_addr, sizeof(serv_addr));
+
+            unsigned char* chp = (unsigned char*) &serv_addr.sin_addr.s_addr;
+            if (5 != sscanf(filename + sizeof(ltcpSrch) - 1,
+                            "%"SCNu8".%"SCNu8".%"SCNu8".%"SCNu8":%"SCNu16, chp + 0, chp + 1, chp + 2,
```
6by9: We already do the equivalent for normal network connections using inet_aton at 1123. That's the normal call used for converting host names to addresses. Any reason we need to roll our own here?

6by9:
> As we only support tcp for listening, and the IP address is unused, then only the port is valid.

The IP address is typically used to bind the listening port to one specific network interface on a multi-homed system, so I'd say it is used. The code appears to read it and populate serv_addr.sin_addr.s_addr with it.

popcornmix:
Okay, how about:
``raspivid -l tcp://0.0.0.0:1234`` for listening for connection
``raspivid -o tcp://192.168.1.2:1234`` for sending

Is there much value in supporting both writing to a file and sending over a network (I'm thinking not much, but others may have different ideas).

host_applications/linux/apps/raspicam/RaspiVid.c
luked99: TEMP_FAILURE_RETRY() is the macro. It's defined in unistd.h if __USE_GNU is defined, which I think we are defining already (via _GNU_SOURCE).

Using it will stop this being easily portable to other glibc versions (e.g. musl) so maybe this should be done in a followup patch?

6by9:
> Okay, how about:
```raspivid -l tcp://0.0.0.0:1234``` for listening for connection
```raspivid -o tcp://192.168.1.2:1234``` for sending

I'm good with that.

> Is there much value in supporting both writing to a file and sending over a network (I'm thinking not much, but others may have different ideas).

Simultaneously? I'd agree not, and it complicates the code quite a lot.
Doing that does mean that the code out to protect against both a -o and -l being specified on the same command-line, or at least do something sensible if both are specified.

**Comment by popcornmix** [comment]
Yes specifying both ``-l`` and ``-o`` should be an error.

**Comment by maroviher** [comment]
I would also add the listening mode for UDP, to do that I have to use sendto instead of fdopen->fwrite. That's why I thought, if I change so much, it could not be a big deal to use a regular file write and network write simultaneously. Some people can find such feature useful, e.g. live observation over network and simultaneously write to a file on a PI.

Some examples one more time:

#wait for an incoming tcp connection on all interfaces.
#After a client is connected, start sending data to a client
#and write it to a local file simultaneously
raspivid -o myfile.h264 -on tcp://0.0.0.0:1234 -l 

#wait for an incoming tcp connection on network interface 
#which has a given IP 192.168.1.2 assigned
#after a client is connected, start sending data to a client
raspivid -on tcp://192.168.1.2:1234 -l

#wait for an incoming udp packet on all interfaces
#after a client is connected, start sending data to a client
raspivid -on udp://0.0.0.0:1234 -l

#start sending UDP packets to a remote IPv4 host 195.12.1.2:1234
#and save the video stream to a local file myfile simultaneously
raspivid -on udp://195.12.1.2:1234 -o myfile

#initiate a TCP connection to a remote IPv4 host 195.12.1.2:1234
#and save video to a file simultaneously
raspivid -on tcp://195.12.1.2:1234 -o myfile

#initiate a TCP connection to a remote IPv4 host 195.12.1.2:1234
#and send video to it
raspivid -on tcp://195.12.1.2:1234


What do you think?

**Comment by maroviher** [comment]
No interest in network listening features?

**Comment by 6by9** [comment]
I have reservations about adding more big new features to Raspivid/still. They were meant as demo code and relatively simple examples of how to communicate with the camera. They're morphing into huge things which have so many bells and whistles that trying to grasp the basics is becoming harder.

While network listening is just modifying the setup that is relatively simple to comprehend. Adding in output to multiple locations puts in yet more stuff to the buffer callbacks, and those are already busy with all the code for inline motion vector, circular buffer, and saving time stamps. 
If raspivid is to get extra features like this, I think we need to sort really simple example apps first (hello_pi type level but using MMAL), and then refactor the buffer callback to break out the various options and make it more readable. These aren't tasks I'm asking you to do, but would benefit from being done.

**Comment by JamesH65** [comment]
Good points. Might have the chance to look at this next year. Some thinking
needed I suspect. Might be worth creating a library so we can have many
small apps that provide the sort of facilities mentioned here but have the
same underlying 'camera' base.

File: host_applications/linux/apps/raspicam/RaspiVid.c (hunk 11 above)
Comment (6by9) [inline host_applications/linux/apps/raspicam/RaspiVid.c]: I was implying that inet_aton (as used at 1123) is the normal way to handle network addresses, rather than rolling your own sscanf line.

**Comment by Ferroin** [comment]
In general I will comment on two things about this:
1. I personally have no significant interest in this other than helping make sure it's done sanely if it's supported (I currently have nothing that uses the camera, and even if I did, I'd need netcat because I would be using SCTP instead of TCP or UDP so that I would have a reliable connection without the insane overhead imposed by TCP).  I see far too many projects that just do stupid things when trying to add networking support.
2. Regarding the evolution of raspi{vid,still}:  Many people don't care that the code was intended to be demo's, it works and it's a lot easier to use tools that already do what you need than to write your own.  That said, I think that splitting the core camera stuff out into a library and then doing some simple command-line tools that let you use that library from a shell script (or languages which don't have bindings) would be a good thing both for clarity and for efficiency.

**Comment by 6by9** [comment]
My current position is that I'd happily accept the current patch to add listening if it used either ```-l tcp://<addr>:<port>``` or ```-l -o tcp://<addr>:<port>``` for the listening option, leaving all other existing options alone. 

Simultaneous streaming and local saving I'll currently push back on due to the significant changes it requires. Part of my reservation is that you really then need to be passing the buffers to separate threads to have the network send and file save happening in parallel, so another big jump in complexity.

Your example of
```
#wait for an incoming udp packet on all interfaces
#after a client is connected, start sending data to a client
raspivid -on udp://0.0.0.0:1234 -l
```
is curious. What are you considering as a UDP connect? Such a thing does not exist as UDP is a connection-less protocol.
The client could send a UDP packet to the server, the server can extract the remote IP address and port, and then start sending data back to that address, but it's not a connect.

maroviher:
Sorry,

 "UDP connect" is poorly expressed. UDP is a connection-less protocol. But sometimes it does matter which of counterparts sends the first packet, for a communication to be possible.

Imagine following:
a RPI is connected to your home Router with a real IPv4 address. You have set up a UDP port forwarding rule to you PI, so that your RPI is accesible from Internet by UDP. You want to stream a video from a RPI to your smartphone. A smartphone is also behind a NAT, but without port forwarding(typical LTE/3G connection). In order to communicate with each other, a first UDP packet must come from a smartphone to a RPI, so that a NAT table of your LTE/3G mobile provider got initialized and the RPI->smartphone way back UDP packets would not be drop in a NAT router of your 3G mobile provider.

Tha is what I meant with:

#wait for an incoming udp packet on all interfaces
#after a client is connected, start sending data to a client
raspivid -on udp://0.0.0.0:1234 -l

maroviher:
I will just add a new -l switch for TCP listening only and add some more error handling (-EINTR). Should I make a new pull request?

host_applications/linux/apps/raspicam/RaspiVid.c
maroviher: sscanf on that place also checks the correctness of a string.
A current version of raspivid will just crash if you type:
raspivid -o tcp://0.0.0.0
Segmentation fault (core dumped)

6by9:
I know the use case you're meaning - effectively you're auto-configuring the remote IP address/port based on whoever last sent data to your port. It does require a chunk more changes that make the code harder to understand. There are things happening in the new year that should allow resourcing of tidying stuff up, so at that point adding UDP "listen" in a tidy manner is feasible.

 Error handling on your patch would be good.
If there's a way to crash the existing code, then a second patch (can be on the same PR) to fix that happily accepted.

6by9:
Can you squash these two commits now?
If one had implemented just TCP listening, and the other just fixing the seg fault on missing port, then having them as two commits makes sense. But as it stands we have an intermediate command line option available of ltcp:// which has the potential to confuse.

I'm happy with the code overall - many thanks for your efforts.

maroviher:
All commits were squashed in a one single commit.

pelwell:
@6by9 The github GUI now allows us to squash and merge or rebase all of the commits attached to a PR. For more complicated scenarios you still have to resort to the command line, but this is a step forward.

6by9:
@maroviher Thank you.
@pelwell Useful to know - thanks.

Any other reviewers want to chime in?

6by9:
Ping @luked99 and @JamesH65. Do you want to review again, or should @popcornmix merge as is?

JamesH65:
I'm OK with it.

6by9:
@popcornmix: happy to merge this? Luke seems to be elsewhere at the moment, but both James and I are happy.

PR #366, opened 2017-01-17
Fix for Issue 365 - Luma mode can result in double sized files

JamesH65:
Implemented a fix provided by 6x9, ensures that zero or small
camera callback buffer lengths are handled correctly

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStillYUV.c
+++ b/host_applications/linux/apps/raspicam/RaspiStillYUV.c
@@ -57,7 +57,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <unistd.h>
 #include <errno.h>
 
-#define VERSION_STRING "v1.3.5"
+#define VERSION_STRING "v1.3.6"
 
 #include "bcm_host.h"
 #include "interface/vcos/vcos.h"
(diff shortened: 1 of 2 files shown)


6by9:
Just a thought, the video port will only produce an EOS on stopping the clock (not supported with MMAL), so I don't think the RaspVidYuv change is needed, but safest to have it in there.

JamesH65:
I did wonder about the change in the Vid version, but think it's safer this
way - if underlying code changes we should ensure things don't break as
much as possible.

popcornmix:
Personally I'd prefer "Fix for Issue 365" is left out of the first line of commit message as it is meaningless to a reader without further cross referencing. Including the module affected is useful. e.g.
"raspicam: Fix for double sized luma files" and move the "Fix for Issue 365" to the description.

Also enabling editing of PRs (from upstream maintainers) is useful (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) as it allows minor tweaks before merging

JamesH65:
I think I have fixed the issues flagged up - I closed the previous PR and
created a new one. I suspect there is a way of updating the PR, but i
didn't find one!

popcornmix:
In general you shouldn't need to create a new PR. Just make your JamesH65:Issue#365 branch look correct and force push and it will update.
I'd still prefer the "raspicam:" prefix to show what area commit applies to.
(Note we also cherry pick these commits into vc4 tree so it is useful to see at a glance this is an arm side application fix, rather than a gpu camera fix).

PR #367, opened 2017-01-17
Fix for Luma mode double sized files in timelapse

JamesH65:
Implemented a fix provided by 6x9, ensures that zero or small
camera callback buffer lengths are handled correctly

Fix for issue 365

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStillYUV.c
+++ b/host_applications/linux/apps/raspicam/RaspiStillYUV.c
@@ -57,7 +57,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <unistd.h>
 #include <errno.h>
 
-#define VERSION_STRING "v1.3.5"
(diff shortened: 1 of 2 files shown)


pelwell:
We can fixup the commit message to include "raspicam:" when it is merged/squashed.

JamesH65:
I've updated my fork of userland to have the prefix. This should mean the
PR is updated...

pelwell:
Indeed. Now just make the header lowercase and you're there...

popcornmix:
> We can fixup the commit message to include "raspicam:" when it is merged/squashed.

Yes. That wouldn't work if there were two (desired) commits for example, so even better is edits aren't needed.

 > This should mean the PR is updated...

Indeed it is.

JamesH65:
Arghghghghghhghghhghg.

popcornmix:
Any chance of:

>  Now just make the header lowercase and you're there...

so we can merge this.

JamesH65:
Soz, thought it had been merged.  I need to tweak the branch a bit as I accidentally added an extra commit.

PR #372, opened 2017-01-26
SIGUSR2 should capture and exit even if verbose is false

AdrianAntunez:
Hi guys,

I recently got merged a pull request #368, I noticed that it only worked if verbose flag is active. I know it was a dummy error and I blame myself for that!! As a result, I provide a fix that allows to use `USR2` signal in both executions (verbose and non-verbose) in that pull request.

Apologise for any inconvenience.

Cheers,

Adrian.


6by9:
Looks fine to me.

PR #374, opened 2017-02-07
Add pkg-config file for mmal

anthonyryan1:
Presently a number of applications use the hardcoded library path for MMAL. I think we would be best to have a pkg-config file for it just like some of the other libraries provided.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ include_directories("${PROJECT_BINARY_DIR}")
 include(FindPkgConfig QUIET)
 if(PKG_CONFIG_FOUND)
 	# Produce a pkg-config file
-	foreach(PCFILE bcm_host.pc  egl.pc  glesv2.pc  vg.pc brcmegl.pc  brcmglesv2.pc  brcmvg.pc )
+	foreach(PCFILE bcm_host.pc egl.pc glesv2.pc vg.pc brcmegl.pc brcmglesv2.pc  brcmvg.pc vcsm.pc mmal.pc )
 		configure_file("pkgconfig/${PCFILE}.in" "${PCFILE}" @ONLY)
 		install(FILES       "${CMAKE_CURRENT_BINARY_DIR}/${PCFILE}"
 			DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig")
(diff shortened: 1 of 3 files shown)


popcornmix:
Seems reasonable. @6by9 any objections?

6by9:
Packaging is not an area I know anything about, but if it achieves the intent then I'm happy with it.

popcornmix:
Same here. But I believe this is intended to create a file called mmal.pc which some people find useful.
And I've tested it and it does create a file called mmal.pc.

The only question is whether the set of libs referenced are complete for "mmal".
Peoples requesting "mmal" package will get ``-lmmal -lmmal_core -lmmal_util -lmmal_vc_client -lbcm_host``. Would adding, say -lvcsm to that list make sense?

anthonyryan1:
Sorry I should have clarified my intention with this a bit better. The pkg-config file is a shortcut that application developers can use to avoid knowing the installation quirks of various libraries in order to use them.

With pkg-config, an application doesn't need to know that mmal is installed in /opt/vc and ffmpeg is installed in /usr/lib, it simply asks pkg-config to include mmal & ffmpeg and the application developer can get about their work.

The only notable user of this so far is recent git master builds of Kodi which actively look for a mmal pkg-config file. I've tested patches for converting ffmpeg and libav as well and verified it makes them easier to install on varying distributions.

As for adding -lvcsm I'm afraid I'm not familiar enough to comment on this, today is only my second day playing with mmal. I can say that we should include everything a person would be expected to link in order to fully use mmal. If there are portions of mmal that would not work without -lvcsm, we should add it. If someone more aware of the scope of mmal's functionality could comment on that I can amend this pull request as needed.

Finally, if we need to make any changes to mmal.pc at a later date, it's pretty painless. Additions to the file aren't likely to cause any regressions.

popcornmix:
@6by9 may be able to offer an opinion on whether vcsm fits in to mmal libs or stands alone.
Kodi does require vcsm now.

6by9:
Anything that uses MMAL zero copy buffers will need vcsm, but it's more a standalone library than part of MMAL itself. 

Add it as a new package which the MMAL package has a dependency on, or include it in the MMAL package and the very few people who want it standalone get MMAL as an overhead? I don't really have any view either way.

anthonyryan1:
I have added a separate pkg-config file for vcsm and made mmal depend upon it.

Please let me know if there are any other changes you would like to see made.

PR #376, opened 2017-02-25
RaspiStill: Fix regression in d35be767 - uninitialised variable

6by9:
d35be767 relied on state->width/height being zero before setting
up the defaults, but the state structure is just on the stack
in main() and not initialised.
Initialise the whole structure to 0.

https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=175445


pelwell:
That change taxed my reviewing powers.

6by9:
> That change taxed my reviewing powers.

Can I make my next patch less than 8 characters changed and yet still be meaningful?! :-)

PR #379, opened 2017-03-03
Add OpenGL library versioning

gazoo74:
Some applications like QtWebEngine require specific library versions.
Make them happy by providing that.

https://wiki.qt.io/RaspberryPi2EGLFS

### DIFF
--- a/interface/khronos/CMakeLists.txt
+++ b/interface/khronos/CMakeLists.txt
@@ -93,3 +93,8 @@ target_link_libraries(brcmWFC brcmEGL)
 target_link_libraries(brcmOpenVG brcmEGL)
 
 install(TARGETS brcmEGL brcmGLESv2 brcmOpenVG brcmWFC DESTINATION lib)
+
+set(DESTDIR \$ENV{DESTDIR})
+include(GNUInstallDirs)
+install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink libbrcmEGL.so ${DESTDIR}/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/libEGL.so.1)")
+install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink libbrcmGLESv2.so ${DESTDIR}/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/libGLESv2.so.2)")


popcornmix:
I tried rebuilding userland libs with this PR and then running an app that used them:
```
/opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/kodi.bin: error while loading shared libraries: libGLESv2.so.2: cannot open shared object file: No such file or directory
```
```
pi@domnfs:~ $ ldd /opt/xbmc-cmake/arm-linux-gnueabihf/lib/kodi/kodi.bin | grep GLES
	libGLESv2.so.2 => not found
pi@domnfs:~ $ ls -l /opt/vc/lib/libGLESv2.*
lrwxrwxrwx 1 root root     14 Mar  4 18:29 /opt/vc/lib/libGLESv2.so -> libGLESv2.so.2
lrwxrwxrwx 1 root root     18 Mar  4 18:29 /opt/vc/lib/libGLESv2.so.2 -> libGLESv2.so.2.0.0
-rw-r--r-- 1 root root 105768 Mar  4 18:29 /opt/vc/lib/libGLESv2.so.2.0.0
pi@domnfs:~ $ objdump -p  /opt/vc/lib/libGLESv2.so |grep SONAME
  SONAME               libGLESv2.so.2
```
I think the problem is the change of SONAME means any code that had been built against the previous libs will no longer run after updating (without a rebuild).

That is quite an obstacle to accepting this PR.

gazoo74:
@popcornmix: I did not find a way to keep the previous SONAME. And btw, I don't think it is what we want.

Did you run ldconfig?

gazoo74:
Here is a new version suggested by @tpetazzoni that will not break existing compiled binaries, and also satisfy the needs for qtwebengine.

gazoo74:
@popcornmix: do you have time for a review on this PR?

popcornmix:
The firmware GL driver is going to be deprectated in the near future, with the arm side driver the preferred option.
We have already added copies of firmware libs called:
libbrcmEGL.so, libbrcmGLESv2.so libbrcmOpenVG.so
which we'd prefer users of the firmware library to switch to.
At some point we will remove libEGL.so, libGLESv2.so libOpenVG.so from /opt/vc/lib to avoid conflicts when both sets of libs are installed.

Due to this, I don't think this PR is going to be helpful going forward.

gazoo74:
@popcornmix: I applied the patch to brcm* libraries. BTW, if you do not want this patch, you can close the PR.

PR #380, opened 2017-03-07
host-apps/dtoverlay: don't install scripts in random location

yann-morin-1998:
Currently, we add a custom command that installs the dtoverlay pre/post
scripts in ${CMAKE_BINARY_DIR}/../../bin

However:

  - this points outside of the package directory; it even points two
    directories higher;

  - when doing cross-compilation, this is definitely not the place where
    the /bin directory really is;

  - the scripts are already properly installed without this code.

Remove that code, it serves no purpose and breaks for cross-compilation.

Signed-off-by: "Yann E. MORIN" <yann.morin.1998@free.fr>

### DIFF
--- a/host_applications/linux/apps/dtoverlay/CMakeLists.txt
+++ b/host_applications/linux/apps/dtoverlay/CMakeLists.txt
@@ -22,12 +22,4 @@ add_custom_command(TARGET dtoverlay POST_BUILD COMMAND ln;-sf;dtoverlay;dtparam)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/dtparam DESTINATION bin)
 
 set(DTOVERLAY_SCRIPTS dtoverlay-pre dtoverlay-post)
-foreach(_script ${DTOVERLAY_SCRIPTS})
-   add_custom_command(
-     TARGET dtoverlay
-     COMMAND ${CMAKE_COMMAND}
-     -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${_script}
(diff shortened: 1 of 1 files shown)


pelwell:
The purpose of these command is to copy the scripts into build/bin during the build phase to make them easier to locate. Obviously they aren't absolutely necessary, but it was nice to be able to see all the build targets in a single location before installing them. Do you know of a cleaner way to achieve the same thing?

yann-morin-1998:
@pelwell Sorry, I don't see the point. What is this "build/" directory you're talking about?

`${CMAKE_BINARY_DIR}/../../bin` poitns _outside_ of the build tree, two levels upper.

For example (cross-compile options and the likes just ignored for the sake of the example):

    $ mkdir ~/foo/bar/buz   # Deep enough
    $ cd ~/foo/bar/buz
    $ git clone https://github.com/raspberrypi/userland/
    $ cd userland
    $ cmake
    $ ls -F ~/foo/bar
    bin/ buz/

The scripts are stored in that new `bin/` directory. This corresponds to nothing usefull, especially when one is doing cross-compilation. Worse, it may even clash with a build-system's own directories, or otherwise existing `bin/` directories (for example, if a build is attempted in `/usr/src/userland`, it would clobber files in `/usr/bin`).

Regards,
Yann E. MORIN.

pelwell:
I was doing some work on the dtoverlay command and remembered this PR. It's merged now - sorry for the delay.

yann-morin-1998:
@pelwell Thanks! Better late than never. ;-)

PR #383, opened 2017-03-22
[Trivial] Update comment for vc_dispmanx_resource_read_data

JinShil:
I think this is copying data *from* VideoCore memory, not *to" VideoCore memory.


JinShil:
`vc_dispmanx_resource_read_data` copies data *from* VideoCore memory *to* `dst_address` and `vc_dispmanx_resource_write_data` copies data *to* VideoCore memory *from* `src_address`.

pelwell:
Yes - the diff was showing the name of the previous function as context.

PR #384, opened 2017-03-23
raspicam updates

6by9:
A couple of updates that I'd forgotten to create the PR for.

1 - zero copy is the potentially contentious one as it adds the dependency on VCSM, but that isn't an issue on Raspbian where it has been present for at least 2 years.
2 - minor uninitialised variable fix.
3 - raspividyuv is copying code across from raspivid
4 - use nproc jobs in the buildme script.


JamesH65:
All looks fine to me.

PR #389, opened 2017-04-16
Escaped the * characters in the imv markdown readme

ChrisSalisbury:
The example split command reads incorrectly due to \*68\* being interpreted as italics in markdown (i.e. it is the equivalent of \<i\>68\<\/i\> in html).


6by9:
Thanks for the correction. Merged.

PR #392, opened 2017-04-21
Make vchiq_test functional test more interesting.

ED6E0F17:
For cosmetic reasons it is nice to have some indication of what "vchiq_test -f" is doing.
The logging points were chosen to narrow down a Linux 4.11 failure with cache flushing.

ref: http://lists.infradead.org/pipermail/linux-rpi-kernel/2017-April/006149.html


pelwell:
No objection from me.

PR #395, opened 2017-05-06
Add support to stream the video both to a file and over stdout (even when capture is not active)

danielealbano:
Hi,

I am working on a personal project involving a raspberry pi zero w and a raspberry pi camera v2.1.

I was planning to capture, save and stream as preview the frames of the camera but after playing a bit with raspivid I discovered that it wasn't able to save and to stream properly.
A lot of people were suggesting to use `tee` to store the video and cvlc to stream it but, one of the reasons I was planning to use raspivid was the ability to start/stop recording and the ability to create a new file on every new start.

After a while I ended up patching the raspivid to be able to do so.

With this patch instead of using tee to duplicate the stdin over a file and the stdout you can use raspivid directly and you can even pause the recording while still having the preview (if the preview over stdout is enabled the MMAL_PARAMETER_CAPTURE is never set to 0 within the while loop but instead is enabled before entering the loop and then disabled when the loop ends).

This is the command I am using
```
./raspivid -w 1280 -h 720 -b 600000 -t 0 -f -e -v -s -sp -so -o /home/pi/test_video.h264 | cvlc -v stream:///dev/stdin --sout '#standard{access=http,mux=ts,dest=:8080' :demux=h264
```

The only part of the code that I modified but not tested is the one related to the circular buffers, I do know which is the use case.

Thanks!

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -139,6 +139,7 @@ typedef struct RASPIVID_STATE_S RASPIVID_STATE;
 typedef struct
 {
    FILE *file_handle;                   /// File handle to write buffer data to.
+   FILE *preview_stdout_file_handle;            /// File handle to write buffer data to.
    RASPIVID_STATE *pstate;              /// pointer to our state in case required in callback
    int abort;                           /// Set to 1 in callback if an error occurs to attempt to abort the capture
    char *cb_buff;                       /// Circular buffer
@@ -178,7 +179,8 @@ struct RASPIVID_STATE_S
    int framerate;                      /// Requested frame rate (fps)
    int intraperiod;                    /// Intra-refresh period (key frame rate)
    int quantisationParameter;          /// Quantisation parameter - quality. Set bitrate 0 and set this for variable bitrate
-   int bInlineHeaders;                  /// Insert inline headers to stream (SPS, PPS)
+   int bInlineHeaders;                 /// Insert inline headers to stream (SPS, PPS)
+   int preview_over_stdout;            /// Output to the stdout
    char *filename;                     /// filename of output file
    int verbose;                        /// !0 if want detailed run information
    int demoMode;                       /// Run app in demo mode
@@ -323,6 +325,7 @@ static void display_valid_parameters(char *app_name);
 #define CommandRaw          32
 #define CommandRawFormat    33
 #define CommandNetListen    34
+#define CommandPreviewOverStdout 35
 
 static COMMAND_LIST cmdline_commands[] =
 {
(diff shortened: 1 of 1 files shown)


6by9:
I can see what you're trying to do, but the code is starting to get very jumbled. At the moment there are a few holes too. 
- What happens if you add ```-o - -so``` to the commandline? Two things writing to stdout :-(
- You're calling encoded data "preview", when preview is normally always on the screen and comes from camera->output[0].
- Does the recording always start with header bytes in all modes? From a quick read I'm not convinced it does. In some modes which don't automatically enable inline headers this will result in an unplayable file.

I think there's a better approach, but it needs a bit more thought first, then time and effort. to implement. 
If we allow multiple -o parameters, we can build up a list of destinations. In the callback, iterate over the list and call a handler for each destination in turn with the new buffer.
Niggles: The -fl (flush) option needs to be allowed multiple times as each destination may or may not want that behaviour. We probably want multiple -pts options as well, possibly dynamically set up -ih (inline headers), and allow some destinations to take the circular buffer contents on trigger whilst others always stream (probably easiest to make the circular buffer always active and destination 0). There's a need for a new option that signals whether the destination should always send encoded data, or only when triggered.

@JamesH65 what are your thoughts? Worth pulling all the file handling code out into a new source file to make the basics more readable and sharable with raspividyuv, and the raw output (-raw & -rf) options?
Making it backward compatible could be fun as currently it won't matter the order of -ih/-fl/-pts with respect to -o, but if we support multiple -o parameters then there will be a dependency. First thought is to normally require that those parameters must come before the -o but apply any trailing ones to the last destination once CLI parsing has completed.

It needs a little thought first, but should be <a day to knock most of it into shape once the preferred solution is decided on.

JamesH65:
@6by9 @danielealbano There is a real need to consolidate the file handling so that all the camera apps can have the same facilities without cut and paste of code. That also applies to the segmented mode stuff and the circular buffer stuff. I think that should be a separate task, once that is done we can start to play about with the sort of extra features described here. The code has got a little out of hand - anything like this will be beneficial. 

With regard to the backwards compatibility, I think it's worth the hit. I doubt many people would fall foul of the ordering, and as long as we document it thoroughly we should be OK. The vast majority won;t see a difference.

danielealbano:
@6by9 having multiple destinations was my first thought but both because of the little time available to do it and both because it would have required a major refactoring I preferred to take the simplest approach possible.

The reason for which I called it "preview" is that the stdout is not affected by the pause, raspivid is always sending data to stdout when -so option is used so you can use. In my case I am using VLC to stream the preview and a simple web page to send signals to raspivid to start/stop the recording and the split option to reopen the destination file when it starts to record.

I know that there are better ways for sure, but this implementation to me seemed like a quick win :)

About your points:

- What happens if you add -o - -so to the commandline? Two things writing to stdout :-(
Indeed this is bad, I will fix it on my branch

- You're calling encoded data "preview", when preview is normally always on the screen and comes from camera->output[0].
I explained the reason about but TLDR the -so is not affected by pause/record state

- Does the recording always start with header bytes in all modes? From a quick read I'm not convinced it does. In some modes which don't automatically enable inline headers this will result in an unplayable file.
I will double check, it should because I "duplicated" all the writes done on file_handle

6by9:
The raspicam apps generally are overdue a refactor and clean up.
Thank you for your efforts, but my feeling is we should do the cleanup rather than accept yet more changes that would be reverted or having to add backwards compatibility. We'll make sure the rework can fulfil your use case, even if not via quite the same command line options.

JamesH65:
I'm with @6by9 on this one - better to refactor in the near future than have to do even more work once something like this is added. 

We'll have a chat in the office tomorrow to see what can be done and when. If nothing else you have prompted us to get this sorted out!

danielealbano:
Thanks for your hard work!

In the meantime I will do the double stdout output fix, may be someone will find it useful :) 

Thanks again!

PR #403, opened 2017-06-09
add rpi2/rpi3 build command

yfang1644:


6by9:
Unless I'm missing something, this isn't adding an rpi2/3 build command. It is making the build solely ARMv7 so it won't work on any Pi A/B/0/0W/CM1 as they are all still ARM 1176jzf. That makes the repo useless to users of all those boards, so that isn't going to happen.
I'm not convinced that it will work if cross-compiling either due to the compiler change. It's also a touch perverse to be modiying arm-linux-gnueabihf.cmake to not use arm-linux-gnueabihf as the compiler.

What is your specific aim for this PR? At the moment there is scant information in the PR or commit text.

JamesH65:
@6by9 Inclined to close this as inactive. Yes?

6by9:
Yes, closing. If @yfang1644 comes back with some more details of what he's trying to achieve then we can reopen.

PR #406, opened 2017-07-09
Add command line option for flicker avoidance

m00fighter:
This change enables a user to run "raspivid" with various flicker avoidance modes. It was previously discussed at "https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=159115", where "6by9" mentioned the existence of the MMAL option and "cpunks" (me) wrote the patch.

An aspect from which this patch might need improving, is how command lines are parsed and where "static XREF_T flicker_avoid_map[]" is located. To avoid duplication, I placed it in "RaspiCamControl.h", but since I don't know architectural plans for command line parsing, I might have placed it inconveniently for others.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.c
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.c
@@ -933,7 +933,7 @@ void raspicamcontrol_set_defaults(RASPICAM_CAMERA_PARAMETERS *params)
    params->videoStabilisation = 0;
    params->exposureCompensation = 0;
    params->exposureMode = MMAL_PARAM_EXPOSUREMODE_AUTO;
-   params->flickerAvoidMode = MMAL_PARAM_FLICKERAVOID_AUTO;
+   params->flickerAvoidMode = MMAL_PARAM_FLICKERAVOID_OFF;
    params->exposureMeterMode = MMAL_PARAM_EXPOSUREMETERINGMODE_AVERAGE;
    params->awbMode = MMAL_PARAM_AWBMODE_AUTO;
    params->imageEffect = MMAL_PARAM_IMAGEFX_NONE;
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.c
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.c
@@ -61,6 +61,18 @@ static XREF_T  exposure_map[] =
 
 static const int exposure_map_size = sizeof(exposure_map) / sizeof(exposure_map[0]);
 
+/// Structure to cross reference flicker avoid strings against the MMAL parameter equivalent
+
+static XREF_T  flicker_avoid_map[] =
+{
+   {"off",           MMAL_PARAM_FLICKERAVOID_OFF},
+   {"auto",          MMAL_PARAM_FLICKERAVOID_AUTO},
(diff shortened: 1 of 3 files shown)


6by9:
As discussed on the thread, anything in RaspiCamControl.c should be automatically available from any of the raspicam apps. The changes in RaspiVid.c shouldn't be necessary.
Otherwise it looks reasonable.

m00fighter:
Thanks, I will try to see if I can alter the code so as to keep modifications to RaspiVid.c to a minimum. Can you tell, if I manage to change the code to achieve that result, should I just change it in my fork, or does the pull request also need updating? (First time using Git and Github to accomplish anything. I've previously only used CVS, SVN and Monotone.)

6by9:
Having cherry-picked the changes to do a quick test, I see we have a option clash - "fl" is already used for "flush" within RaspiVid. Running "raspivid --help" also shows up "flicker" twice.

I think the above explains why you needed to mod RaspiVid too - Raspivid will have found a match for "fl" in its list of command line options and then never passed the request on to ```raspicamcontrol_parse_cmdline```.  Your option is before "flush" in the table, so you've actually broken selecting flush via "-fl".
Raspistill has no such issue for "fl" matching a command, so it will pass it in to ```raspicamcontrol_parse_cmdline``` to handle.

Add your new change to your local branch as a new commit, and then do another push to github.
The PR is a request to merge the top of your branch into this repo, so it will get automatically updated with the new commit. We can squash all of your commits together in to one commit when we merge the change.

6by9:
Sorry, further observation and change request - the old default behaviour was that flicker avoidance was off, and that ought to be retained rather than adopting "auto" as the default.
Changing the behaviour is avoided so that someone updating their device doesn't suddenly find things breaking/not behaving as before.

m00fighter:
Thanks, all very good points - I had totally missed that "fl" doubled for "flush" and not changing default behaviour is wise. :) I hope to implement the changes on my side by tomorrow.

m00fighter:
I have now gone throught the checklist I made, and I think all the issues you found are addressed. I would appreciate if you could check. And thanks for finding all of them! :)

6by9:
Looks pretty good. Thanks for your efforts.

FYI you can use ```git rebase -i``` to list and edit all your commits that haven't been pushed to the source branch (ie userland/master in this case). They will all be listed as "pick" by default, which means retain the commit as it is. As is shown in the editor that comes up, you can replace pick with edit, squash, drop (or just remove the line), or a couple of others. 
I've just taken your commmits and squashed them into one (which is what will happen when this PR is merged), and there are still a couple of minor changes to RaspiVid that we don't really want. To be slightly picky, there are also a few issues with whitespace being left around. For reference, https://github.com/6by9/userland/tree/flicker_avoid is what the resulting tree looks like.

@JamesH65 I think I'm happy to accept as is, but could you do a quick review too please?

m00fighter:
Thanks, I'll try to be careful about needless white space modifications in future. I think I did cause some "noise" of that sort, by editing code and then reversing edits. I'll then carry on writing my video analysis tool and if there's anything else I can contribute, I'd ask about it. :)

JamesH65:
Apart from the whitespace issue, looks fine to me.

6by9:
Merged. Thanks for your contribution.

PR #418, opened 2017-08-20
pkgconfig: commit to new library naming scheme

psyke83:
Recent firmware images have deprecated the original-named libraries
(libEGL, libGLESv1_CM, libGLESv2, libOpenVG). This means that the
corresponding dummy .pc files will present issues if installed.

Instead, use the proper library names via symlink.

These files will only be detected when the variable
"PKG_CONFIG_PATH=/opt/vc/lib/pkgconfig" is set manually.

--
This commit is related to this issue: https://github.com/raspberrypi/firmware/issues/857

### DIFF
--- a/pkgconfig/egl.pc.in
+++ b/pkgconfig/egl.pc.in
@@ -1,13 +1,1 @@
-prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=${prefix}
-libdir=${exec_prefix}/lib
-includedir=${prefix}/include
-
-Name: EGL
-Description: Fake EGL package for RPi
-Version: 10
-Requires: bcm_host
-Libs: -L${libdir} -lEGL -lGLESv2 -lbcm_host -lvchostif
-Cflags: -I${includedir} -I${includedir}/interface/vmcs_host/linux \
-        -I${includedir}/interface/vcos/pthreads
-
+brcmegl.pc.in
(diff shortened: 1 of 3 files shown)


psyke83:
I go into more details re: the intent of this PR here: https://github.com/raspberrypi/firmware/issues/857#issuecomment-323736009

The existing glesv2, egl and vg pkgconfig files no longer work as expected due to the libraries being removed from Debian stretch's firmware package. By removing these files and creating symlinks to the brcm versions, it ensures that the new library names are used IFF the PKG_CONFIG_PATH is set.

Using the pkgconfig files before applying this PR:
pi@retropie:~ $ PKG_CONFIG_PATH=/opt/vc/lib/pkgconfig pkg-config --libs glesv2
-L/opt/vc/lib -lGLESv2 -lbcm_host -lvcos -lvchiq_arm -pthread

This breaks the build due to /opt/vc/lib/libGLESv2.so being deprecated.

With the PR:
pi@retropie:~ $ PKG_CONFIG_PATH=/opt/vc/lib/pkgconfig pkg-config --libs glesv2
-L/opt/vc/lib -lbrcmGLESv2 -lbcm_host -lvcos -lvchiq_arm -pthread

This PR won't interfere with the standard libraries by default (as long as the pkgconfig remains in a nonstandard search path, i.e. /opt/vc/lib/pkgconfig):
pi@retropie:~ $ PKG_CONFIG_PATH=/opt/vc/lib/pkgconfig pkg-config --libs glesv2
-L/opt/vc/lib -lbrcmGLESv2 -lbcm_host -lvcos -lvchiq_arm -pthread
pi@retropie:~ $ pkg-config --libs glesv2
-lGLESv2

psyke83:
This PR was aimed towards Raspbian. Will close this, as the upcoming firmware packages will include only the non-conflicting files, so that's a good compromise. Thanks.

PR #419, opened 2017-08-30
VCSM: Import DMABUF support

6by9:


6by9:
Closing as the main VCSM changes have come in via the firmware tree. The gitignore change I may repush later, but otherwise it'll remain local.

PR #423, opened 2017-09-19
dtoverlay: no longer require user "pi"

teslaTrooper0:
also $(id -u) seems more reliable than $EUID which can be overwritten by the user

### DIFF
--- a/host_applications/linux/apps/dtoverlay/dtoverlay-post
+++ b/host_applications/linux/apps/dtoverlay/dtoverlay-post
@@ -4,7 +4,7 @@ if [ "$DISPLAY" == "" ]; then
 fi
 CMD="which lxpanelctl >/dev/null 2>&1 && lxpanelctl alsastart >/dev/null"
 if [ $(id -u) -eq 0 ]; then
-        exec su $(id -un) -c "$CMD"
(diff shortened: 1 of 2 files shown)


pelwell:
I'm happy with the `$EUID` -> `$(id -u)` substitution, but not `pi` -> `$(id -un)`. By my reading, if `$(id -u)` is `0` then `$(id -un)` is `root`.

The purpose of that logic is to (attempt to) ensure that the hook scripts run with the same user id as the desktop environment, otherwise some of it won't work. The code as it was has the advantage of working if the desktop runs as `pi`, whereas the new version looks like it won't work at all.

If @spl237 knows of a better way to establish the id of the current user of the desktop then I'd be happy to improve the compatibility.

teslaTrooper0:
@pelwell This should run it as the logged in user, ignoring sudo or su.

pelwell:
Have you tested that these updated scripts select user `pi` when run from the GUI when the audio controls are changed?

Ruffio:
@teslaTrooper0 have you tested @pelwell suggestion?

teslaTrooper0:
I honestly don't remember but [this change](https://github.com/raspberrypi/userland/commit/d574b51a60a075baefe863670466ee24e6c4256e) solved the issue either way.

PR #426, opened 2017-09-27
mmal_wrapper: Fix a short allocation for pools and queues

Terminus-IMRC:
There is a bug in MMAL wrapper code that is because of a short allocation for pools and queues. The bug corrupts stack, so some programs exit with error.

Program to reproduce: https://github.com/Terminus-IMRC/mmal-wrapper-issue-short-allocation

Please merge this to your repo.

Thanks.

### DIFF
--- a/interface/mmal/util/mmal_component_wrapper.c
+++ b/interface/mmal/util/mmal_component_wrapper.c
@@ -151,7 +151,7 @@ MMAL_STATUS_T mmal_wrapper_create(MMAL_WRAPPER_T **ctx, const char *name)
    if (status != MMAL_SUCCESS)
       return status;
 
-   extra_size = (component->input_num + component->output_num * 2) * sizeof(*private);
+   extra_size = (component->input_num * sizeof(MMAL_POOL_T*)) + (component->output_num * (sizeof(MMAL_POOL_T*) + sizeof(MMAL_QUEUE_T*)));
    private = vcos_calloc(1, sizeof(*private) + extra_size, "mmal wrapper");
    if (!private)
    {
--- a/interface/mmal/util/mmal_component_wrapper.c
+++ b/interface/mmal/util/mmal_component_wrapper.c
@@ -151,7 +151,7 @@ MMAL_STATUS_T mmal_wrapper_create(MMAL_WRAPPER_T **ctx, const char *name)
    if (status != MMAL_SUCCESS)
       return status;
 
-   extra_size = (component->input_num + component->output_num) * 2;
+   extra_size = (component->input_num + component->output_num * 2) * sizeof(*private);
    private = vcos_calloc(1, sizeof(*private) + extra_size, "mmal wrapper");
    if (!private)
    {


interface/mmal/util/mmal_component_wrapper.c
```diff
@@ -151,7 +151,7 @@ MMAL_STATUS_T mmal_wrapper_create(MMAL_WRAPPER_T **ctx, const char *name)
    if (status != MMAL_SUCCESS)
       return status;
 
-   extra_size = (component->input_num + component->output_num) * 2;
+   extra_size = (component->input_num + component->output_num * 2) * sizeof(*private);
```
6by9: I'll agree it is wrong as written, but it's not * sizeof(*private).
Looking at line 176 it's using each extra entry only as a pointer to either a MMAL_POOL_T, or a MMAL_QUEUE_T. There's a MMAL_POOL_T per input port, and both a MMAL_POOL_T and MMAL_QUEUE_T per output port. So I'd say something along the lines of 
```
extra_size = (component->input_num * sizeof(MMAL_POOL_T*)) + (component->output_num * (sizeof(MMAL_POOL_T*) + sizeof(MMAL_QUEUE_T*))
```
would be the desired value.

I've never used these wrapper functions (I'm not sure anyone did in anger), so there may well be gremlins lurking.

Terminus-IMRC: You are missing the last `)`, but, yeah, your formula is better to understand (the resultant value is the same as mine since sizeof(*some pointer*) is always 4). I've pushed this change.

The change just increases a memory to allocate, so no problems are going to occur it seems.

6by9: (I think I'm right on this, but could be hideously embarrassing myself!)

My main point was that you hadn't done ```sizeof(some pointer)``` though, you'd done ```sizeof(*(some pointer))```, ie the sizeof the thing pointed to, or ```sizeof(MMAL_WRAPPER_PRIVATE_T)```. That is != 4. (76 I think - I haven't got code to hand to invoke it).

```sizeof(private)``` I would have accepted (slightly grudgingly), as it is just a pointer and would be size 4, but having looked into what it is using that for then it's far better to be specific..

6by9:
@JamesH65 I'm happy with the commits once squashed (the intermediate one is still wrong).
Do you want to review?

interface/mmal/util/mmal_component_wrapper.c
Terminus-IMRC: Oh I missed it. Thanks!

6by9:
(Sorry for the delay on this one - James has taken a long weekend and didn't review this before he went. Should be addressed Tuesday).

Terminus-IMRC:
No probs. Thanks!

JamesH65:
Looks OK to me.

6by9:
Thanks both. Now merged.

PR #445, opened 2017-12-29
user-vcsm: Add an interface to get a VideoCore address of handle

Terminus-IMRC:
This ioctl call was defined in the kernel header, but not implemented to user-vcsm. This call is really useful for using QPU etc.

### DIFF
--- a/host_applications/linux/libs/sm/user-vcsm.c
+++ b/host_applications/linux/libs/sm/user-vcsm.c
@@ -693,6 +693,59 @@ unsigned int vcsm_vc_hdl_from_hdl( unsigned int handle )
 }
 
 
+/* Retrieves a videocore (bus) address from a opaque handle
+** pointer.
+**
+** Returns:        0 on error
+**                 a non-zero videocore address on success.
+*/
+unsigned int vcsm_vc_addr_from_hdl( unsigned int handle )
+{
+   int rc;
+   struct vmcs_sm_ioctl_map map;
+
+   if ( (vcsm_handle == VCSM_INVALID_HANDLE) || (handle == 0) )
+   {
+      vcos_log_error( "[%s]: [%d]: invalid device or handle!",
+                      __func__,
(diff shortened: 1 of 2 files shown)


6by9:
No objections from me.
I must admit to going the roundabout route before by allocating via vcsm, getting the VC handle, and then locking via the mailbox interface (https://github.com/raspberrypi/firmware/wiki/Mailbox-property-interface#lock-memory) to give me the bus address. Doing it directly via vcsm is nicer.

Terminus-IMRC:
> Doing it directly via vcsm is nicer.

I think so too! In addition, vcsm supports cache cleaning/invalidating which may enable cache-enabled CPU-QPU shared memory. I'm writing a userland library for that.

6by9:
Please be aware that there is a rewrite of vcsm in progress to use CMA for the allocation, export as a dmabuf, and hand it to the VPU/QPU for use. There shouldn't be any real change to the external API of the userland library, but the whole kernel ABI is being redone.

Terminus-IMRC:
Thanks for your information. It is being developed on a *public* repository/branch? I've done some changes on the current vcsm about cache flushing which is almost ready for PR, so I'd like to know where it is.

6by9:
It's not in a fit state for any form of release as yet, but is my main project at the moment.
The intent is that there will be a new device under /dev, and vcsm_init will try that first, and fall back to the existing /dev/vcsm if it is not accessible. All userland vcsm calls should have equivalent functionality between the two.

Create your PR and as long as it is acceptable then I'll port it to the new system too.

Terminus-IMRC:
Thanks! I see.

popcornmix:
As @6by9 says this can be done through mailbox interface, but supporting directly from vcsm makes sense to me, so no objections.

6by9:
Thanks, merged.

Terminus-IMRC:
Thanks!

Issue #400, opened 2017-06-05
About exposure mode and ISO


tpengti:
When setting exposure=off, adjusting ISO is invalid  (adjusting the shutter seems to be ineffective).
ISO is valid only when exposure mode is on.

What is the relationship between ISO and analog gain and digital gain, I have to modify the userland, add the analog gain and the digital gain of the parameter adjustment, but the prompt function is not implemented.
Whether you can open the Gamma parameter?

Looking forward to your reply, thank you very much.

6by9:
> When setting exposure=off, adjusting ISO is invalid 

Correct. ISO setup is actually done as a slightly modified auto exposure mode with fixed gain settings.
You should be able to set a manual shutter speed even if exposure mode auto and ISO specified. Only ISO 100, 200, 400, and 800 are supported.

> What is the relationship between ISO and analog gain and digital gain

ISO = analogue gain * digital gain * 100 / gain_for_iso_100.
On OV5647 gain_for_iso_100 is x1. On IMX219 they it was actually calibrated and is x1.84.

> I have to modify the userland, add the analog gain and the digital gain of the parameter adjustment, but the prompt function is not implemented.

No, as discussed on the forums numerous times, there is no direct user control of analogue and digital gains. I had looked in to it and it proved pretty awkward to achieve.

> Whether you can open the Gamma parameter?

Unlikely in the main camera stack.
There are the discussions on the forums over capturing the raw Bayer images and processing those directly - https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=109137.
There is also then the ISP component which is still alpha but should eventually allow you greater control of parameters, however you as the user also then have to take full control over AWB, AE, AGC, lens shading, gamma, etc. If you want the lowest level of control, then that is probably the route you should be investigating.

tpengti:
thank you for your reply
Why manual exposure mode, can not adjust ISO?

6by9:
> Why manual exposure mode, can not adjust ISO?

It's jsut the way it evolved.
Consider exposure=off to be a lock of all current settings, therefore nothing will change.

Setting ```MMAL_PARAM_EXPOSUREMODE_AUTO``` and ```MMAL_PARAMETER_SHUTTER_SPEED``` to a non-zero value will disable automatic exposure control whilst leaving the automatic gain control active. You can then set your desired ISO value via ```MMAL_PARAMETER_ISO```.

tpengti:
@6by9 
[...]

PR #446, opened 2018-01-08
Add range validation to ISO

emilbayes:
I don't know that this range is set in stone, but this is the range that @6by9 mentions is valid in this issue: https://github.com/raspberrypi/userland/issues/400

Also given that issue, should those specific values be strictly validated or is any int in-between valid?

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiCamControl.c
+++ b/host_applications/linux/apps/raspicam/RaspiCamControl.c
@@ -182,7 +182,7 @@ static COMMAND_LIST  cmdline_commands[] =
    {CommandContrast,    "-contrast",  "co", "Set image contrast (-100 to 100)",  1},
    {CommandBrightness,  "-brightness","br", "Set image brightness (0 to 100)",  1},
    {CommandSaturation,  "-saturation","sa", "Set image saturation (-100 to 100)", 1},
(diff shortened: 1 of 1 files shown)


emilbayes:
Although, we have successfully used ISO values of `50`, but anything less seems to give exactly the same image (highly controlled environment)

6by9:
It's complicated.
ISO is meant to be a calibrated value. On OV5647 (V1.3 camera) ISO 100 = analogue gain of x1.0. On IMX219 (V2.1 camera) ISO 100 = analogue gain of x1.84, and therefore an ISO value lower than 100 can be achieved.
It's further complicated in that during preview/video modes there needs to be some flexibility to allow correct reproduction of a scene whilst also allowing things such as flicker avoidance to set the exposure time to a multiple of the mains flicker period. That requires some level of adjustment of the gain to compensate for fixed steps on the exposure time. To achieve that the tuning has fixed values for nominated ISO values of 100, 200, 400, and 800.
For stills mode, there is no such restriction, and therefore any value of ISO will be accepted and should be actioned.

TBH If you want fixed gains, then you're better off using the -ag and -dg commands to explicitly set analogue and digital gains.

JamesH65:
Given @6by9 comments, I think this PR can be closed?

6by9:
Yes, will close. The range checking is going to give potentially odd results, and the comment only goes along with that.

PR #451, opened 2018-01-29
Verbose is not disabled when capture is streamed

ledazul:
Verbose was being disabled when a capture is sent to stdout, but verbose is sent to stderr, so there is no need to disable it.


6by9:
Thank you.
If being fussy then most coding styles would retain the braces around both clauses in the if/else if either of them needed it, but I won't be that strict here.

ledazul:
Thanks for the tip, I'll take it into account from now on.

Issue #391, opened 2017-04-21
Add bcm_host library support for arm64


raidensnake:
I ran the recent userland on debian arm64 and it was missing the required bcm_host library needed to compile retropie's rpi optimised SDL libraries. I was able to compile the library manually in the source folder with no issues. Maybe it would help by adding that to the cmake lists so it can be compiled in arm64.

6by9:
Create a pull request with your proposed change and it will be reviewed.

We're not prioritising arm64 development but are happy to work with others who wish to try to get it working. There are various places where interfacing with the GPU firmware may be "fun" as that is only 32bit, so any pointers in shared structs are going to mismatch in size.

raidensnake:
what I'm saying is the bcm_host library compiles as is on arm64. but it has to be done manually.

stiltr:
I needed bcm_host for libcec on ARM64 and was able to compile it by removing the exclusion from host_applications/linux/CMakeLists.txt. Everything _seems_ to work ok from what I can tell (libcec works fine). Is there a reason this is being excluded on ARM64? I'm happy to create a PR if needed, but "it worked for me" isn't exactly "fully tested".

JamesH65:
@stilr A PR enables us to look at the code quickly to see what will change and check for any obvious issues, so that would be a good start.

raidensnake:
Since I don't really know any of that stuff all I did was go into the host_applications/linux/libs/bcm_host manually and use cmake to compile it.

luked99:
On 4 February 2018 at 16:42, raidensnake <notifications@github.com> wrote:
>
> Since I don't really know any of that stuff all I did was go into the host_applications/linux/libs/bcm_host manually and use cmake to compile it.


Isn't the problem that VideoCore exposes a 32 bit API, where all the
handles are 32 bit. So although you might be able to just recompile,
it won't actually work when you pass a 64 bit pointer across and
expect it to come back in one piece.

I think various schemes have been suggested for fixing this, but
no-one has invested the time to turn it into a pull request.

Or perhaps it's more complicated than that.

6by9:
> Isn't the problem that VideoCore exposes a 32 bit API, where all the handles are 32 bit. So although you might be able to just recompile, it won't actually work when you pass a 64 bit pointer across and expect it to come back in one piece.
I think various schemes have been suggested for fixing this, but no-one has invested the time to turn it into a pull request.
Or perhaps it's more complicated than that.

That pretty much sums it up.

[...]

PR #455, opened 2018-02-05
Add bcm_host to ARM64 builds

stiltr:
See #391

### DIFF
--- a/host_applications/linux/CMakeLists.txt
+++ b/host_applications/linux/CMakeLists.txt
@@ -1,8 +1,6 @@
 # linux apps
 
-if(NOT ARM64)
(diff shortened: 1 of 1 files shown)


pvizeli:
Will that be merged soon?

popcornmix:
As mentioned in #391 having the vchiq services that bcm_host can invoke actually work correctly on arm64 is a much bigger issue.
But no real reason to refuse to build this when it compiles, so we can merge it.

ED6E0F17:
gles/egl is just a thin wrapper on top of the firmware, so hello_triangle is working here, but I know there are serious issues with porting some of the other APIs to 64bit.

PR #457, opened 2018-03-15
pkgconfig: vcsm: Add -lvcos to Libs

Terminus-IMRC:
`libvcsm.so` is linked with vcos in https://github.com/raspberrypi/userland/blob/master/host_applications/linux/libs/sm/CMakeLists.txt#L15 , so changed to specify it in pkgconfig file.


6by9:
Looks OK to me, but I'm no expert on packaging.

Terminus-IMRC:
Thank you for commenting every time! I greatly appreciate it.

popcornmix:
Not an expert either, but it looks plausible. Thanks.

Issue #178, opened 2014-06-02
Build issue causing raspivid to fail (components not registered)


ajlennon:
libmmal_vc_client.so makes use of **attribute**(constructor) to ensure that supplier components (e.g. camera) are loaded when the static library is loaded.

raspivid, and possibly other applications, link against libmmal_vc_client.so, causing the ctor to execute, but there is no needed dependency.

Some build environments (e.g. Yocto/OpenEmbedded) pass the '--no-as-needed' linker flag which removes the dependency on libmmal_vc_client and thus needed components are not registered.

In this situation raspivid then gives an error of the form

root@raspberrypi:~# raspivid -o test
mmal: mmal_component_create_core: could not find component 'vc.ril.camera'
mmal: Failed to create camera component
mmal: main: Failed to create camera component
mmal: Failed to run camera app. Please check for firmware updates

For further details see  https://lists.yoctoproject.org/pipermail/yocto/2014-June/019933.html

Ruffio:
@ajlennon is this still an issue?

YamSoup:
yes

ajlennon:
@Ruffio not sure I'm afraid. I submitted a patch to meta-raspberrypi to prevent the --as-needed being used. I I get a chance I'll see if that is working

PeterDaveHello:
+1

Ruffio:
@ajlennon is it working?

JamesH65:
Anyone contributing to the thread have any further comment or possible fixes?

PeterDaveHello:
Sorry I don't, I'm not using this for a while.

6by9:
> so, is this a configuration issue of the build environment?

Yes, as covered in some of the linked issues.
If the linker adds the ```--no-as-needed``` flag, then it is over-eager in stripping out code from libmmal_vc_client.so to the extent of having no VideoCore components. Ideally there needs to be something extra to signal to the linker that this code is required, but it's not been a priority to resolve.

6by9:
@pelwell from https://github.com/raspberrypi/userland/issues/303#issuecomment-204397114 where you point a similar VPU issue. Having had a play I don't believe that is going to help here.
As I understand it, the issue here is that the linker when processing raspistill/vid has gone through and determined that we don't call any of the libmmal_vc_client.so functions, therefore it isn't needed as a dependency to the app. It misses that the constructor will do stuff that will make a difference.

[...]

PR #459, opened 2018-03-20
host_apps: Add --no-as-needed to stop libmmal_vc_client being dropped

6by9:
Some versions of GCC have --as-needed turned on by default.
The binding of libmmal_vc_client is such that the constructor
registers the supported components with the MMAL core, and nothing
calls into it directly. The linker can't tell this, decides
it is unused, and promptly drops it as a dependency - cue no
VideoCore components.

Adding --no-as-needed means that the linker leaves it alone.
It's not nice, but there doesn't appear to be a better solution.

See #178.

### DIFF
--- a/host_applications/android/apps/vidtex/CMakeLists.txt
+++ b/host_applications/android/apps/vidtex/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 2.8)
 
 SET(COMPILE_DEFINITIONS -Werror -Wall)
+
+# Set --no-as-needed to stop the linker discarding mmal_vc_client
+# as it can't see that the constructor registers a load of functionality
+# with the MMAL core.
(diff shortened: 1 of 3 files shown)


6by9:
@popcornmix Seeing as this has been merged to the vc4 tree independently, are you happy for this to be merged to bring them back into track?

popcornmix:
Does this exactly match firmware version? If so then merge.

6by9:
It's effectively a squash of 0bd679a63cb9a19e670adbb8b064eaf21e164a32 and 
eb336ff17cfeb20c07e6ea6dde0067de6578467c in vc4. 
I think you cherry-picked the original version of this PR to your vc4 working branch and pushed it. eb336f was a fixup of that.

I'm happy to cherry-pick the two individual commits and update this PR if you'd prefer.

PR #467, opened 2018-05-14
khronos: backport EGLuint64KHR typedef from recent Khronos

hhromic:
The newer `gstreamer1.0-plugins-base` package version `1.14` uses `EGL_EXT_image_dma_buf_import`,
which expects the `EGLuint64KHR` typedef that is present in recent versions of Khronos.
However, the older version included in userland does not provide it.

This patch backports the missing typedef from recent Khronos into userland.
See: <https://www.khronos.org/registry/EGL/api/EGL/eglext.h>

### DIFF
--- a/interface/khronos/include/EGL/eglext.h
+++ b/interface/khronos/include/EGL/eglext.h
@@ -190,6 +190,10 @@ typedef EGLBoolean (EGLAPIENTRYP PFNEGLSIGNALSYNCKHRPROC) (EGLDisplay dpy, EGLSy
 typedef EGLBoolean (EGLAPIENTRYP PFNEGLGETSYNCATTRIBKHRPROC) (EGLDisplay dpy, EGLSyncKHR sync, EGLint attribute, EGLint *value);
 #endif
 
+#ifndef EGL_KHR_uint64_typedef
+#define EGL_KHR_uint64_typedef 1
+typedef khronos_uint64_t EGLuint64KHR;
+#endif /* EGL_KHR_uint64_typedef */
 
 #ifdef __cplusplus
 }


JamesH65:
@popcornmix @6by9 This looks fairly benign to me, any thoughts on merging?

6by9:
My concern is that then people expect it to work.
> gstreamer1.0-plugins-base package version 1.14 uses EGL_EXT_image_dma_buf_import, which expects the EGLuint64KHR typedef 

What does it use it for? Has anyone sanity checked that it will never go through the code path that is going to use a potentially undefined and unsupported type?

Checking the GStreamer repos, they've already patched it. https://github.com/GStreamer/gst-plugins-base/blob/master/gst-libs/gst/gl/egl/gsteglimage.c#L89
https://github.com/GStreamer/gst-plugins-base/commit/2f3ff84f4082097a36732a91734121fb209efb84

hhromic:
Hi, @6by9 , thanks for pointing to the upstream fix. When I submitted this PR, that fix wasn't in order yet.
I made the PR to address specifically Gstreamer, and now that it's fixed in there this PR/patch is irrelevant if nothing else benefits from it. Let me know if I should close it now.
Thanks @JamesH65 for the review too.

JamesH65:
I'd be inclined to close. The gstreamer change doesn't quite seem the appropriate fix, but taking this PR  now will probably cause more grief than help.

kraj:
there are other apps e.g. weston also expecting this define ```EGLuint64KHR``` to be made available as well.

hhromic:
looks like this PR should be resurrected and a proper solution, i.e. update Khronos?, be discussed.

ndufresne:
> What does it use it for? Has anyone sanity checked that it will never go through the code path that is going to use a potentially undefined and unsupported type?

Yes, we do check if the DMABuf extensions are provided by the stack before using it in GStreamer. It's a distro implementation, so we try and have as much as possible built-in and enabled bits at run-time. One can explicitly out-compile it when building for the pre-mesa RPi GL stack of course.

PR #470, opened 2018-07-01
RaspiVid: add sliced encoding

shizukachan:
Add interface to sliced encoding feature of VideoCore.

Implementation rounds up on (height/16)%slices != 0.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -2120,7 +2120,8 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
    if (state->encoding == MMAL_ENCODING_H264 &&
        state->slices > 1)
    {
-      int slice_row_mb = (state->height/16)/state->slices + (state->height/16)%state->slices?1:0; //round up
+      int frame_mb_rows = (state->height/16) + (state->height%16)?1:0;
+      int slice_row_mb = frame_mb_rows/state->slices + frame_mb_rows%state->slices?1:0; //round up
       MMAL_PARAMETER_UINT32_T param = {{ MMAL_PARAMETER_MB_ROWS_PER_SLICE, sizeof(param)}, slice_row_mb};
       status = mmal_port_parameter_set(encoder_output, &param.hdr);
       if (status != MMAL_SUCCESS)
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -233,6 +233,7 @@ struct RASPIVID_STATE_S
    int64_t lasttime;
 
    bool netListen;
+   int slices;
 };
 
 
@@ -323,6 +324,7 @@ static void display_valid_parameters(char *app_name);
 #define CommandRaw          32
 #define CommandRawFormat    33
 #define CommandNetListen    34
+#define CommandSlices       35
 
 static COMMAND_LIST cmdline_commands[] =
 {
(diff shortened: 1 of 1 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2106,6 +2117,19 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
       }
    }
 
+   if (state->encoding == MMAL_ENCODING_H264 &&
+       state->slices > 1)
+   {
+      int slice_row_mb = (state->height/16)/state->slices + (state->height/16)%state->slices?1:0; //round up
```
6by9: I assume you're trying to ensure that you round the height up such that it divides by slice_row_mb to give exactly state->slices slices of data. I don't think this calculation works correctly though.
Pass a height of 130 and 2 slices in, and I think you'll get 3 slices out each time (slice_mb_rows=4 for 64 lines per slice). The same applies to heights of 129-143.
I know these weird rounding conditions are often nasty, which is why I'm not immediately making a suggestion of a fix.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -440,7 +443,7 @@ static void default_status(RASPIVID_STATE *state)
    state->save_pts = 0;
 
    state->netListen = false;
-
+   state->slices = 1;
```
6by9: The structure is memset to 0, and your check later is that slices>1, therefore you don't need to initialise this to 1.

host_applications/linux/apps/raspicam/RaspiVid.c
shizukachan: Ooops, I forgot about non mod16 video sizes.

The usual method would probably be to round up (video height / 16) to get macroblock rows per frame, then round up (macroblock rows per frame / slice count) to get macroblock rows per slice. This should ensure we never have more slices than specified.

host_applications/linux/apps/raspicam/RaspiVid.c
shizukachan: I believe it's better to set slices to 1 by default even if it has no effect:
1) 0 as a number of slices a video stream has does not make sense. If someone uses state->slices after this, they would have to special case that 0 actually means 1 slice.
2) it's consistent with the style that is already there (why should slices be excluded from initialization?)
3) it documents that the default behavior is to encode with one, full-height slice.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2120,7 +2120,8 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
    if (state->encoding == MMAL_ENCODING_H264 &&
        state->slices > 1)
    {
-      int slice_row_mb = (state->height/16)/state->slices + (state->height/16)%state->slices?1:0; //round up
+      int frame_mb_rows = (state->height/16) + (state->height%16)?1:0;
+      int slice_row_mb = frame_mb_rows/state->slices + frame_mb_rows%state->slices?1:0; //round up
```
6by9: Sorry, I still think that fails the desired criteria.
Height 128 and 2 slices sets slice_row_mb to 5, or 160 lines per slice. In fact many of the heights come out wrong. I'm just making up a simple spreadsheet that runs through the calculation for heights 128-180, prints the result, and then multiplies it back up. Heights 146-176 all end up with a slice_row_mb of 6 (192), when only over 160 requires 6.

I'll have a think about the logic myself - the modulo stuff feels very wrong, but I also appreciate that it may be required for handling 3 slices/frame.

shizukachan: It's possible without the modulo.

6by9:
@JamesH65 Do you fancy applying the final polish here?

JamesH65:
This patch does not appear to work on certain resolution - most noticeably 1080p - I'm getting a firmware death. I'm taking a look but cannot spend a huge amount of time on it.
Example failure :
`raspivid -sl 4 -0 fred.h264`

JamesH65:
This pR has been superceeded by #486 which has now been merged.

PR #471, opened 2018-07-02
glxx_client: Fix glMapBufferOES with GL_STREAM_DRAW

### LINKED ISSUE #246: Cannot glMapBufferOES with GL_STREAM_DRAW
In raspberrypi/userland/interface/khronos/glxx/glxx_client.c
Line 385 & 397
... If glBufferData() is called with GL_STREAM_DRAW then the buffer's cached size is set to zero.

In raspberrypi/userland/interface/khronos/ext/gl_oes_map_buffer.c
Line 66 & Line 94
... If the cached size is zero, then the GL_INVALID_OPERATION is raised and the buffer is not mapped.

The net effect being that you can't call glMapBufferOES on a buffer that was set up using GL_STREAM_DRAW.  I think it ought to work fine though, and it seems like a small bug on glxx_client.c line 385.

timg236:
glMapBufferOES fails if the usage hint passed to glBufferData was
GL_STREAM_DRAW because glBufferData does not initialise the cached copy
of the buffer size.

The usage hint doesn't make any functional difference with these
functions and GL_STREAM_DRAW was added in GLES 2.0 so this looks like
an old check which was never updated.

Fixes #246

### DIFF
--- a/interface/khronos/glxx/glxx_client.c
+++ b/interface/khronos/glxx/glxx_client.c
@@ -382,7 +382,7 @@ GL_API void GL_APIENTRY glBufferData (GLenum target, GLsizeiptr size, const GLvo
       {
          if( ((target == GL_ARRAY_BUFFER && state->bound_buffer.array != 0) ||
               (target == GL_ELEMENT_ARRAY_BUFFER && state->bound_buffer.element_array != 0)) &&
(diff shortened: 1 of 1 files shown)


JamesH65:
LGTM, @6by9 any comment prior to merge?

6by9:
I haven't the foggiest idea, but looks plausible.

PR #472, opened 2018-07-02
glxx_client: Fix glGetBufferParameteriv GL_BUFFER_SIZE

### LINKED ISSUE #323: Requesting buffer size messes up client state
When `glGetBufferParameteriv` is called with `GL_BUFFER_SIZE`, it creates and stores a `GLXX_BUFFER_INFO_T` but sets `id` and `cached_size` only. `mapped_pointer` and `mapped_size` get random values. This makes every subsequent call to `glBufferData` fail and `glGetBufferPointervOES` also returns false information from this point on.

timg236:
glGetBufferParameteriv updates the client side cached copy of the
buffer_size. However, when it updated the cache it by calling
glxx_buffer_info_set it trashed the mapped_pointer and mapped_size
values because they are in the same structure as the cached_size but
this was being allocated on the stack.

The trampled mapped_pointer / mapped_size could cause crashes or invalid
buffer data to be uploaded.

Change to read the current cached value and then modify just the
cached_size preserving whatever was in the cache before.

Fixes #323

Signed-off-by: Tim Gover <tim.gover@raspberrypi.org>

### DIFF
--- a/interface/khronos/glxx/glxx_client.c
+++ b/interface/khronos/glxx/glxx_client.c
@@ -2107,6 +2107,7 @@ GL_API void GL_APIENTRY glGetBufferParameteriv (GLenum target, GLenum pname, GLi
          {
             GLXX_CLIENT_STATE_T *state = GLXX_GET_CLIENT_STATE(thread);
             GLXX_BUFFER_INFO_T buffer;
+            glxx_buffer_info_get(state, target, &buffer);
             buffer.cached_size = params[0];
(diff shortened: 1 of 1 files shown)


timg236:
Tested by modifying hello_triangle2/triangle2.c to map the buffer via glMapBufferOES, read the buffer size via glGetBufferParameteriv and unmapped buffer and verified that the correct vertex data was uploaded.

JamesH65:
@popcornmix Lgtm, any further comment prior to merge?

PR #475, opened 2018-07-08
Possible mistype in mmal_metadata.h

zeme-wana:
Possible mistype MMAL_META(T)DATA ...

### DIFF
--- a/interface/mmal/mmal_metadata.h
+++ b/interface/mmal/mmal_metadata.h
@@ -41,14 +41,14 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /* @} */
 
 /** Generic metadata type. All metadata structures need to begin with these fields. */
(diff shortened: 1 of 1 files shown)


popcornmix:
@6by9 These are never used in the code so I guess it doesn't really matter.
What do you think? Rename/leave/delete?

pelwell:
Rename or delete.

6by9:
Agreed - delete.

It was intended as an equivalent to the VC_IMAGE metadata structures. Seeing as there doesn't even appear to be an implementation of the function (only a prototype), get rid of it. There is a reference to it in mmal.h and CMakelists.txt (so it gets installed), so those out to be cleaned up too.

popcornmix:
@6by9 do you mean delete the mmal_metadata.h file and the current include of it from mmal.h?

6by9:
Yes, delete mmal_metadata.h and the references to it in mmal.h and CMakelists.txt

Issue #429, opened 2017-10-04
Auto exposure issue with timelapse capture


tomsaul:
When using the `-t <duration> -tl <frequency>` options to capture a sequence of images (with or without -bm) the application is not waiting to get an appropriate determination of the exposure to use.  With -bm it simply uses the first value (typically very over exposed), and without it it generally slowly converges on a proper exposure as more frames are taken.  I suspect there should be a sleep (perhaps with the duration indicated by another setting) to allow the exposure to be determined after the camera is activated, but before images are saved.  This is done in some of the other capture methods - but not for timelapse!

6by9:
So what value have you specified as duration and frequency?
The code does wait for ```frequency``` milliseconds before taking the first capture.

I've just run ```raspistill -w 640 -h 480 -o foo%d.jpg -t 30000 -tl 2000 -bm, -set```, and it takes captures starting at 2 seconds, which is more than enough time for AE to settle. It always starts at exposure of 10ms and gain x1.0, and in two runs it's adapted to 1.4msec and x1.0 when pointed out the window, or 32.9ms and x1.3 gain looking in the office.

tomsaul:
So far anything below about -tl 1000 results in this problem.  The faster the rate, the more pronounced it is.  The point is that this capture should not start before the hardware is ready, so the rate specified should not be a factor.  The other cases in the code seem to add a fixed delay (1000 as I recall) to allow time for this process, but that was skipped on timelapse mode.  Another option would be to change the initial capture to something like `max(<tl interval>, <init delay>)`, where `<init delay>` could be the new setting, or the hard coded value used elsewhere!

tomsaul:
And FWIW - the exact same issue occurs with the raspivid program.  The image initially displays overexposed, and slowly corrects.

JamesH65:
I would expect value below 1000 to start to become unreliable. The systems need time to start up the camera, receive the first frame, dump it, tyhen start to receive good frames, then get start its AE/AGC loops. That can  take upto 1000ms (more if you are setting large shutter times).

tomsaul:
This is the crux of the problem report.  Most other modes of capture have a fixed startup delay to allow this to occur.  The Timelapse mode in raspicam does not, and while I have not looked at the code in raspivid, my guess is that it also is not doing this.  In both cases, all the image control options are using their defaults.

JamesH65:
[...]

Issue #473, opened 2018-07-04
raspistill --timelaps waits an initial delay before starting


pstray:
Seems raspistill --timelaps waits the timelapse period before taking the first period. I would expect it to snap a picture immediately and then every timelapse period therafter.

PR #476, opened 2018-07-18
Fixed up initial timelapse delay

JamesH65:
When in timelapse mode, the delay before the first shot
was the timelapse value. However, if a low value is set
this does not give enough time for the AE etc to settle.

In addition, the first image was only taken after the
first delay, rather than immediately which is what would
be expected. So a 5minute timelapse delay meant the 1st
image was only taken after 5 minutes.

The patch sets the delay for the first image capture to
be a constant value, 1000ms. So always time for the AE etc
to settle.

Fixes: #429, #473


6by9:
Looks reasonable to me.

PR #477, opened 2018-07-19
Raspicam tweaks

6by9:
Support for TGA and PPM encoders to Raspistill

Bug fix to RaspiVidYuv to shut down the correct output port before closing file handles.


JamesH65:
LGTM

PR #479, opened 2018-07-20
Segment timestamp

JamesH65:
Option to specify timestamp in the segment filename
    
In segmented mode, this allows you to specify that the file
name for each segment uses a timestamp, not a segment number.
Activated using %s instead of %d in the filename.


JamesH65:
CLOCK_REALTIME appears to be wall clock time, and I think is adjusted as such i.e. uses timezones.

kukabu:
CLOCK_REALTIME returns UTC time

JamesH65:
OK, l have updated significantly as suggested by 6by9, now used strftime to format the string which gives endless combinaison of exciting filenames. However, you don't get the formatters %d or %u as they are used to detect if you want segment numbers.

6by9:
I'm good with that.

PR #481, opened 2018-07-25
Raspicam annotate updates and hello_encode fixes

6by9:


JamesH65:
Not actually tried it but LGTM, although the mixed in whitespace changes made it a bit confusing to read. I assume there is a matching vc4 firmware change - what happens with older firmware, does it fall back or just not annotate anything?

6by9:
If you look at the commits instead of the combined changes, then they are clean patches.
Yes, there's a firmware patch incoming. Merged to our tree but not released via rpi-update. Old firmware will just ignore the 3 new fields as they are on the end (done that way deliberately!)

JamesH65:
Cool, so it's a yes from me!

PR #486, opened 2018-09-14
RFC: H264 slices

JamesH65:


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2538,12 +2572,21 @@ int main(int argc, const char **argv)
...
-   {
       fprintf(stderr, "\n%s Camera App %s\n\n", basename(argv[0]), VERSION_STRING);
-      dump_status(&state);
+
+   if (state.slices > 1 && state.width > 1280)
+   {
+      if (state.verbose)
+         fprintf(stderr, "Warning: slices reduced to 1 as width > 1280\n\n");
+
+      state.slices = 1;
```
6by9: Technically not required as you check that width <=1280 before actually setting it.

JamesH65: It was simply to give the user an idea why setting slices might not be doing what they think when verbose is on. Without this, when the parameters are dumped it says slice = whatever, even though they may have been completely ignored later.

PR #489, opened 2018-10-05
MMAL & RaspiStillYuv updates

6by9:
Two unrelated changes.

RaspiStillYuv supported RGB24 but not BGR24. Add it.

Cherry-picking a firmware source commit to userland. Allows flags buffers to be passed without requiring a buffer to also be allocated as long as alloc_size is set to 0.


6by9:
@JamesH65 Any chance of a review?

JamesH65:
LGTM

6by9:
I'll take that as approval to merge then

PR #491, opened 2018-10-11
Fix for bug where %04d no longer works as a valid output file name

comeradealexi:
It appears when timestamps were added, certain combinations of output filename no longer work. Specifying %04d would take the code down the timestamp route.

I've not been able to test this code works on linux but I believe it should.

It now works with %04d and %d while preserving the new timestamp feature.

I brought up this issue here where 6by9 recommended I make a patch :)

https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=224569

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -59,6 +59,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
 #include <memory.h>
 #include <sysexits.h>
 
@@ -1089,11 +1090,21 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
    {
       // Create a new filename string
 
-      // If %d or %u, assume a segment numnber, otherwise use the formatter as
-      // input to strnftime
-      if (strstr(filename,"%u") != NULL || strstr(filename,"%d") != NULL)
-      {
-          asprintf(&tempname, filename, pState->segmentNumber);
+	  //If %d/%u or any valid combination e.g. %04d is specified, assume segment number.
+	  bool bSegmentNumber = false;
+	  const char* pPercent = strstr(filename, "%");
+	  if (pPercent)
+	  {
+	  	pPercent++;
+	  	while (*pPercent != '\0' && isdigit(*pPercent))
+	  		pPercent++;
+	  	if (*pPercent == 'u' || *pPercent == 'd')
+	  		bSegmentNumber = true;
+	  }
+
+	  if (bSegmentNumber)
+	  {
+		asprintf(&tempname, filename, pState->segmentNumber);
       }
       else
       {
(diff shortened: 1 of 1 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1089,11 +1090,21 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
       // Create a new filename string
 
-      // If %d or %u, assume a segment numnber, otherwise use the formatter as
-      // input to strnftime
-      if (strstr(filename,"%u") != NULL || strstr(filename,"%d") != NULL)
-      {
-          asprintf(&tempname, filename, pState->segmentNumber);
+	  //If %d/%u or any valid combination e.g. %04d is specified, assume segment number.
+	  bool bSegmentNumber = false;
+	  const char* pPercent = strstr(filename, "%");
```
6by9: ```strchr(filename, '%');``` would be ever so slightly more efficient as it's only looking to match a single char rather than the string.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1089,11 +1090,21 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
...
+	  if (pPercent)
+	  {
+	  	pPercent++;
+	  	while (*pPercent != '\0' && isdigit(*pPercent))
```
6by9: My logic says that '\0' isn't a digit, therefore the ```isdigit``` condition would abort the loop anyway at '\0'.
Whether a compiler agrees with me is another question :-o

comeradealexi: Good point! https://en.cppreference.com/w/cpp/string/byte/isdigit Suggests isdigit would behave the way you describe.

host_applications/linux/apps/raspicam/RaspiVid.c
comeradealexi: Nice! I will make this change :)

comeradealexi:
@6by9 I've made the optimisations you've recommended :)

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1092,11 +1092,11 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
 
 	  //If %d/%u or any valid combination e.g. %04d is specified, assume segment number.
 	  bool bSegmentNumber = false;
-	  const char* pPercent = strstr(filename, "%");
+	  const char* pPercent = strchr(filename, "%");
```
6by9: Please retest. Needs to be single quotes around the %, not double (single char vs a string)

comeradealexi:
Thanks 6by9! I've updated again. Do I need to do anything else now to get it merged?

6by9:
@JamesH65 quick review from you?

We do still have two different email addresses used on the commits - one for @AlexHoughton and one for @comeradealexi. Whilst they would appear to be both you, do you wish to choose which one is used?

comeradealexi:
@comeradealexi please! I wasn't aware I had two different email addresses... I'll double check my PC at home to see if it's got a different email associated or something!

6by9:
OK, it may need some manual magic to rework it.

```
git config --global user.email
git config --global user.name
```
will confirm your settings, although you can override per project (```git config --get user.email``` instead).
```
git config --global user.email "email@example.com"
git config --global user.name "My Name"
```
to set.

6by9:
I can't amend author details through Github, so have pulled the change down, squashed and altered the author line, and repushed as #492. I also cleaned up a couple of whitespace issues that had crept in, and added some commit text.

Closing this PR as it has been replaced.

PR #492, opened 2018-10-12
Raspivid: Fix for bug where %04d no longer works.

6by9:
When inserting the time/date stamp into the filename was added
certain combinations of output filename no longer work. As it
was only searching for "%u" or "%d". Specifying %04d would take
the code down the timestamp route but with invalid parameters.

Make the search slightly more intelligent to allow specifying
the number of digits to insert in the filename.


6by9:
@JamesH65 
Replaces #491 as I don't want to push direct to the repo, and can't use the GUI to amend authors (squash will always take the author from the first commit, and that was the one that was wrong).

PR #495, opened 2018-10-31
Delay after first frame, burst-mode #494

spink-al:
https://github.com/raspberrypi/userland/issues/494


popcornmix:
@6by9 good to merge?

6by9:
Yes, I'm happy.

PR #497, opened 2018-11-14
raspivid: Fix so you can record raw without encoding too.

6by9:
You could use raspividyuv to do it, but there's no reason to
prevent doing it in raspivid.

Just to make @gsh happy.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -2916,11 +2916,12 @@ int main(int argc, const char **argv)
          {
             // Only encode stuff if we have a filename and it opened
             // Note we use the copy in the callback, as the call back MIGHT change the file handle
-            if (state.callback_data.file_handle)
+            if (state.callback_data.file_handle || state.callback_data.raw_file_handle)
             {
                int running = 1;
 
                // Send all the buffers to the encoder output port
+               if (state.callback_data.file_handle)
                {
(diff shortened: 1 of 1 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2937,7 +2938,7 @@ int main(int argc, const char **argv)
                }
 
                // Send all the buffers to the splitter output port
-               if (state.raw_output) {
+               if (state.callback_data.raw_file_handle) {
                   int num = mmal_queue_length(state.splitter_pool->queue);
```
JamesH65: Is the raw_output flag still requried elsewhere?

6by9: I don't believe so.
It's used for setup, and nuked if the file handle fails to open. The rest uses the file handle.

6by9:
@JamesH65 Any further comments?

host_applications/linux/apps/raspicam/RaspiVid.c
JamesH65: Should probably remove it then, but TBH, hardly worth the effort. Othewise LGTM.

6by9: Sorry, misintepreted your comment.
The command line parser sets raw_output. The setup code then checks it and will open the file and assign callback_data.raw_file_handle.
The rest of the code then only looks at callback_data.raw_file_handle.

PR #503, opened 2018-11-29
Moved all camera 'setting' option code to single source

JamesH65:
All the cam apps were duplicating the same code for outputing camera
settings (e.g. exposure), so have moved it to the CamControl file
so we now only have one set.

Also fixed up a couple of missing bad error returns that were not being
translated correctly from the MMAL error to ints.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStill.c
+++ b/host_applications/linux/apps/raspicam/RaspiStill.c
@@ -47,7 +47,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 // We use some GNU extensions (asprintf, basename)
-#define _GNU_SOURCE
+#ifndef _GNU_SOURCE
+   #define _GNU_SOURCE
+#endif
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


host_applications/linux/apps/raspicam/RaspiStill.c
```diff
@@ -216,7 +215,6 @@ static void store_exif_tag(RASPISTILL_STATE *state, const char *exif_tag);
 #define CommandSignal       16
 #define CommandGL           17
 #define CommandGLCapture    18
-#define CommandSettings     19
```
6by9: Does it cause a problem having a gap in the numbers? Do we rewrite this as an enum definition?

JamesH65: No problem with gaps, but I'll change to enums in a different PR anyway. unless I remember why I did that way originally...

JamesH65:
I did wonder about calling it 'default_camera_control_callback`, or having a small shim, perhaps make the name change anyway, even if I don't add a shim, which just seems like adding code for codes sake.

6by9:
To shim, or not to shim, that is the question!
I'm happy not to have one if you make it ```default_camera_control_callback```, as at least it is then obvious what is going on.

JamesH65:
OK, changed name - also fixed some build warnings as I was there.

PR #504, opened 2018-11-29
Default runs that require interaction to infinite time

JamesH65:
Things like keyboard and signal mode by default will only last 5s
which is confusing, so now they default to run indefinitely
unless overidden with the -t option.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStill.c
+++ b/host_applications/linux/apps/raspicam/RaspiStill.c
@@ -48,7 +48,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // We use some GNU extensions (asprintf, basename)
 #ifndef _GNU_SOURCE
-#define _GNU_SOURCE
+   #define _GNU_SOURCE
 #endif
 
 #include <stdio.h>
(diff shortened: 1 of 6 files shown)


6by9:
It feels a touch hacky having to have yet another state variable.
(a) It's not used beyond the command line parsing, so could be a local to that function instead of in the state.
(b) You could set timeout to -1 (an invalid time) by default, and check AFTER all the parsing whether it has been changed from that.
(c) it's in camel case!

Minor observation though not from this patch - timeout is defined as an ```int``` in the state structure, but read with ```sscanf("%u")```. We love consistency.

JamesH65:
Good ideas, will fix up the PR. And fix that scanf as well.

host_applications/linux/apps/raspicam/RaspiStill.c
```diff
@@ -47,7 +47,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
 // We use some GNU extensions (asprintf, basename)
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
```
6by9: Going to be a conflict here against #503. Please merge 503 first as it has the nicer indentation.

PR #505, opened 2018-11-30
aarch64: build pass

jesset:
With `-DARM64=ON ` , build pass on arm64 arch.


6by9:
Your commit text is totally lacking in any explanation of what you're changing and why.

I think you're changing buildme to make it work on a Pi running arm64. If so why modify the clause intended for building natively on a non-Pi machine (yes, userland used to build on x86/x86_64). Having to add the ```--native``` option makes no sense and it would be more logical to extend the top "Native compile on the Raspberry Pi" clause.

PR #506, opened 2018-11-30
Converted all the command ID's to enums, code tidy

JamesH65:
Changed the command defgines to enum
Removed any remaining tabs and replaced with spaces.


JamesH65:
Damn, completely forgot about those. Will add them. There's some other defines elsewhere I may do as well, just to get them all out out of the way.

JamesH65:
@6by9 One last review?

6by9:
Not looked too closely, but looks good.
Changing all that whitespace makes reviewing it awkward :-/

6by9:
Separate commit in the PR would have been fine. It might be worth doing a mega-blitz on whitespace over the whole of raspicam in one hit, and then at least future patches should be clean.

PR #509, opened 2018-12-03
Reformat of source to userland coding standard

JamesH65:
used

astyle -A1 -s3 -w

Not applied to non-RPI sourced code.

No functional changes.


6by9:
Looks reasonable.
I'm happy enough with the switch/case format of
```
   switch(foo)
   {
   case BAR:
      //Stuff
     break;
   }
```
although we've then got a couple of funnies with
```
   switch(foo)
   {
   case BAR:
   {
      int foo_bar;
      //Stuff
     break;
   }
   }
```
which just look weird. I tend to indent the case, but I know the kernel coding standard doesn't. Not going to argue it.

JamesH65:
Perfectly possibe to indent the case automagically in astyle, but above is what I tend to do to avoid too much indentation.

pelwell:
I've done the double brace at the same level trick - it looks a bit weird, but better than the alternatives IMO.

PR #510, opened 2018-12-05
CamApps : Removed version number and replace with commit id

JamesH65:
Changed CMakeLists.txt for the camera apps to grab the head commit
hash and pass it to the compiler as a define. Use this to version
the apps rather than the never updated version number.

### DIFF
--- a/host_applications/linux/apps/raspicam/CMakeLists.txt
+++ b/host_applications/linux/apps/raspicam/CMakeLists.txt
@@ -12,6 +12,23 @@ include_directories(${PROJECT_SOURCE_DIR}/host_applications/linux/libs/bcm_host/
 include_directories(${PROJECT_SOURCE_DIR}/host_applications/linux/apps/raspicam/)
 include_directories(${PROJECT_SOURCE_DIR}/host_applications/linux/libs/sm)
 
+# Find the commit hash of the build and pass to the compiler
+execute_process(
+  COMMAND git log -1 --abbrev=12 --format=%h
+  OUTPUT_VARIABLE GIT_COMMIT_ID
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+add_definitions("-DGIT_COMMIT_ID=\"${GIT_COMMIT_ID}\"")
+
+# Determine if we are tainted
+execute_process(
+  COMMAND bash "-c" "git ls-files -m | wc -l"
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  OUTPUT_VARIABLE GIT_TAINTED
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+add_definitions("-DGIT_TAINTED=${GIT_TAINTED}")
+
 set (GL_SCENE_SOURCES
    gl_scenes/models.c
    gl_scenes/mirror.c
--- a/host_applications/linux/apps/raspicam/RaspiHelpers.c
+++ b/host_applications/linux/apps/raspicam/RaspiHelpers.c
@@ -46,11 +46,29 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "RaspiCamControl.h"
 #include "RaspiCommonSettings.h"
 
+#ifndef GIT_COMMIT_ID
+#define GIT_COMMIT_ID "Not found"
+#endif
+
+#if (GIT_TAINTED > 0)
+#define TAINTED " Tainted"
+#else
+#define TAINTED ""
+#endif
+
 static const char *app_name;
+
+void print_app_details(FILE *fd)
+{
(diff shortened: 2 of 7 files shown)


JamesH65:
This will presumably need the build for the apps to be done from the userland repo to make the hash most useful - not sure this is how it is done at the moment though.

pelwell:
Ah ha - we have a CMake expert in our midst!

This looks plausible. Assuming it actually works (i.e. the commit ID extraction and quoting for the printf()s) then I approve.

6by9:
Check with @XECDesign as to how Raspbian will build this and does it get the right hash.

Does it actually behave if you download a snapshot tar/zip of the repo and build that? I've had a number of issues raised recently where I've asked what hash they're using to find they've downloaded a random snapshot.

XECDesign:
Currently, everything is pulled in from the `firmware` when appropriate.

6by9:
> Currently, everything is pulled in from the firmware when appropriate.

Hmm, that could be entertaining as I suspect it'll get the hash of the firmware repo, not the hash of the userland repo which is what we really want.

XECDesign:
Since the firmware repo just has pre-built binaries, it shouldn't.

6by9:
> Since the firmware repo just has pre-built binaries, it shouldn't.

My bad - I was thinking of hello_pi which is copied as source.
As long as we don't try applying this code to that then we're probably OK.

XECDesign:
At some point we should move the libs and binaries out of /opt and provide proper packages (the way we're doing it right now is awful from a Debian packaging point of view). It may cause some issues when that happens, but nothing that we won't be able to work around.

JamesH65:
@XECDesign Worth creating a new issue for that, if there isn't one already. Sounds like a sensible idea that should be implemented sooner rather than later. I cannot see the issues being too onerous.

JamesH65:
Everyone happy to merge? I have tested all the apps, and definitely get the right commit ID .

pelwell:
@6by9's concern about a non-git download needs addressing.

JamesH65:
OK, will test that, although that was why I put in the #ifndef GIT_COMMIT_ID clause.

JamesH65:
Although thinking about it, that won't work, I suspect you would get something like "Not a git repo" in the define rather than the commit ID. Will try it out.

pelwell:
I was hoping that GitHub might add a (for example) .version file containing the head commit ID, but I can't find anything like that.

JamesH65:
A non-git downloaded build gives this...

`raspistill Camera App (commit )`

So, builds OK, just nothing in the ID.

popcornmix:
I've just tested with a firmware build and ``/opt/vc/bin/raspicam`` reports:
``raspivid Camera App (commit e5896a07ad5a)``
which is a commit hash from the internal vc4 repo rather than the userland repo.

JamesH65:
Ah, bummer.

pelwell:
Either we switch to building the utilities from userland or we live with the discrepancy - the information is still useful, just marginally harder to interpret.

6by9:
Insert the build date in the string? At least that is useful to the general user as much as us.

popcornmix:
You might also want to consider added a ``(clean)``/``(tainted)`` flag (see ``interface/vcos/vcos_genversion``) as we do when building firmware.
It uses ``git status --porcelain`` to check if there are local edits to the tree it's been build from.

JamesH65:
Might use something like `git ls-files -m | wc -l` which returns 0-n where n is the number of modified files.

JamesH65:
OK, have also added the tainted flag as suggested by Dom.

However, because of changes I am making in the camera apps that conflict with this stuff, going to close this without merging, and  make another PR once the other stuff is done. Would appreciate any comments on this in the meantime.

popcornmix:
No need for another PR. Once the other changes have been merged, just rebase this tree and force push and all will be good.

JamesH65:
Rebased, debugged, and I think ready to go.

pelwell:
Looks good to me.

6by9:
One thought - this is in the cmake rules, so does it get triggered if you just "make" from the right directory? I don't know where this gets inserted into the process.

6by9:
To answer my own question, it is only triggered when you call cmake, not on make.
I've built tainted code which doesn't declare itself as such, and vice versa.

JamesH65:
I always do my build using the buildme script, which uses cmake, so nowt sure what other things might do.

pelwell:
Ditto. Do we support any other method?

6by9:
I was only thinking that we have the relevant Makefile stuff in vc4 (interface/vcos/vcos_genversion), so was there benefit of not requiring the cmake step every time.
Parts of userland (eg hello_pi) can be built via make alone. For the rest you need to cmake once, but the rest of the time you can just make.
I'm not overly fussed either way, but thought I'd raise it.

PR #514, opened 2018-12-10
Added annotate options to raspiyuv and raspividyuv

JamesH65:
This brings them in to line with the non-YUV variants.


host_applications/linux/apps/raspicam/RaspiVidYUV.c
```diff
@@ -727,6 +728,7 @@ static void camera_buffer_callback(MMAL_PORT_T *port, MMAL_BUFFER_HEADER_T *buff
    {
       int bytes_written = 0;
       int bytes_to_write = buffer->length;
+      int64_t current_time = vcos_getmicrosecs64()/1000;
```
6by9: I'm guessing this is copied from raspivid verbatim, but it seems a touch odd to divide by 1000 here, and then divide again by 1000 when it comes to use the value later. Why not divide by 1000000 here and have done with it?

JamesH65: Indeed, straight copy paste, but will sort it out.

JamesH65:
Happy to merge?

PR #517, opened 2018-12-11
Add GPS annotation to the raspivid output

JamesH65:
This involved:
Moving GPS code to own file so can be used in all apps
Adding option in raspivid to apply GPS data to annotation
  Enabled with the -gps flag on command line plus -a 2 to turn on
  application annotation.

### DIFF
--- a/host_applications/linux/apps/raspicam/CMakeLists.txt
+++ b/host_applications/linux/apps/raspicam/CMakeLists.txt
@@ -43,9 +43,11 @@ set (COMMON_SOURCES
    RaspiCLI.c
    RaspiPreview.c
    RaspiCommonSettings.c
-   RaspiHelpers.c)
+   RaspiHelpers.c
+   RaspiGPS.c
+   libgps_loader.c)
 
-add_executable(raspistill ${COMMON_SOURCES} RaspiStill.c  RaspiTex.c RaspiTexUtil.c tga.c ${GL_SCENE_SOURCES} libgps_loader.c)
+add_executable(raspistill ${COMMON_SOURCES} RaspiStill.c  RaspiTex.c RaspiTexUtil.c tga.c ${GL_SCENE_SOURCES} )
 add_executable(raspiyuv   ${COMMON_SOURCES} RaspiStillYUV.c)
 add_executable(raspivid   ${COMMON_SOURCES} RaspiVid.c)
 add_executable(raspividyuv  ${COMMON_SOURCES} RaspiVidYUV.c)
--- a/host_applications/linux/apps/raspicam/RaspiGPS.c
+++ b/host_applications/linux/apps/raspicam/RaspiGPS.c
@@ -1,0 +1,241 @@
+/*
+Copyright (c) 2018, Raspberry Pi (Trading) Ltd.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the copyright holder nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <memory.h>
+#include <errno.h>
+#include <math.h>
+
+#include "RaspiGPS.h"
+
+typedef struct
+{
+   pthread_mutex_t gps_cache_mutex;
+   gpsd_info gpsd;
+   struct gps_data_t gpsdata_cache;
+   time_t last_valid_time;
+   pthread_t gps_reader_thread;
+   int terminated;
+   int gps_reader_thread_ok;
+} GPS_READER_DATA;
+
+static GPS_READER_DATA gps_reader_data;
+
+#define GPS_CACHE_EXPIRY      5 // in seconds
+
+struct gps_data_t *raspi_gps_lock()
+{
+   pthread_mutex_lock(&gps_reader_data.gps_cache_mutex);
+   return &gps_reader_data.gpsdata_cache;
+}
+
+void raspi_gps_unlock()
+{
+   pthread_mutex_unlock(&gps_reader_data.gps_cache_mutex);
+}
+
+static void *gps_reader_process(void *gps_reader_data_ptr)
+{
+   while (!gps_reader_data.terminated)
+   {
+      int ret = 0;
+      int gps_valid = 0;
+
+      gps_reader_data.gpsd.gpsdata.set = 0;
+      gps_reader_data.gpsd.gpsdata.fix.mode = 0;
+      if (connect_gpsd(&gps_reader_data.gpsd) < 0 ||
+            (ret = read_gps_data_once(&gps_reader_data.gpsd)) < 0 )
+         break;
+
+      if (ret > 0 && gps_reader_data.gpsd.gpsdata.online)
+      {
+         if (gps_reader_data.gpsd.gpsdata.fix.mode >= MODE_2D)
+         {
+            // we have GPS fix, copy fresh data to cache
+            gps_valid = 1;
+            time(&gps_reader_data.last_valid_time);
+            pthread_mutex_lock(&gps_reader_data.gps_cache_mutex);
+            memcpy(&gps_reader_data.gpsdata_cache, &gps_reader_data.gpsd.gpsdata,
+                   sizeof(struct gps_data_t));
+            pthread_mutex_unlock(&gps_reader_data.gps_cache_mutex);
+         }
+      }
+      if (!gps_valid)
+      {
+         time_t now;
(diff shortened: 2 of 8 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1139,14 +1140,76 @@ static void update_annotation_data(RASPIVID_STATE *state)
...
+         if (gpsdata->set & TIME_SET)
+         {
+            time_t rawtime;
+            struct tm *timeinfo;
+            rawtime = gpsdata->fix.time;
+            timeinfo = localtime(&rawtime);
+            strftime(datetime, sizeof(datetime), "%Y:%m:%d %H:%M:%S", timeinfo);
+         }
+
+
```
pelwell: Extra blank line.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1139,14 +1140,76 @@ static void update_annotation_data(RASPIVID_STATE *state)
...
+            {
+               if (isnan(gpsdata->fix.latitude) == 0)
+               {
+                  snprintf(lat, sizeof(lat), "%.6lf", gpsdata->fix.latitude);
+               }
+               if (isnan(gpsdata->fix.longitude) == 0)
+               {
+                  snprintf(lon, sizeof(lon), "%.6lf", gpsdata->fix.longitude);
+               }
+            }
```
pelwell: If I were writing this it would be !isnan and no braces, but as I've been reminded, this isn't the kernel (which means the braces are in the right place).

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2408,6 +2471,11 @@ int main(int argc, const char **argv)
 
    check_camera_model(state.common_settings.cameraNum);
 
+   if (state.common_settings.gps)
+   {
```
pelwell: Braces are used here, but not in the following change. I'd lose them.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+   gps_reader_data.terminated = 1;
+
+   if (gps_reader_data.gps_reader_thread_ok)
+   {
+      if (verbose)
+         fprintf(stderr, "Waiting for GPS reader thread to terminate\n");
+
+      pthread_join(gps_reader_data.gps_reader_thread, NULL);
+   }
+   if (verbose && (gps_reader_data.gpsd.gpsd_connected))
```
pelwell: I see they've been copied from the original but the extra parentheses just make it look more complex than it is.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+   {
+      int ret = 0;
+      gps_reader_data.gpsd.gpsdata.set = 0;
+      gps_reader_data.gpsd.gpsdata.fix.mode = 0;
+      if ((connect_gpsd(&gps_reader_data.gpsd) < 0) ||
+            ((ret = read_gps_data_once(&gps_reader_data.gpsd)) < 0))
+         break;
+
+      int gps_valid = 0;
+      if ((ret > 0) && (gps_reader_data.gpsd.gpsdata.online))
```
pelwell: More extra parens.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+   while (!gps_reader_data.terminated)
+   {
+      int ret = 0;
+      gps_reader_data.gpsd.gpsdata.set = 0;
+      gps_reader_data.gpsd.gpsdata.fix.mode = 0;
+      if ((connect_gpsd(&gps_reader_data.gpsd) < 0) ||
+            ((ret = read_gps_data_once(&gps_reader_data.gpsd)) < 0))
+         break;
+
+      int gps_valid = 0;
```
pelwell: We're not supposed to use C++ style declarations...

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+      gps_reader_data.terminated = 1;
+      raspi_gps_shutdown(verbose);
+      exit(EX_SOFTWARE);
+      return;
+   }
+
+   gps_reader_data.gps_reader_thread_ok = 1;
+   return;
+}
+
```
pelwell: KILL THE TRAILING BLANK LINE!!!11!??

host_applications/linux/apps/raspicam/RaspiGPS.c
JamesH65: This was a C&P from the previous code in Raspistill, which came from a third party, so was relucatant tochange anything in it, but will do anyway!

host_applications/linux/apps/raspicam/RaspiGPS.c
JamesH65: Arghghghghghg

host_applications/linux/apps/raspicam/RaspiVid.c
JamesH65: Have removed some but not all, for reasons of readability rather than consistency.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+      fprintf(stderr, "Closing gpsd connection\n\n");
+
+   disconnect_gpsd(&gps_reader_data.gpsd);
+
+   libgps_unload(&gps_reader_data.gpsd);
+
+   pthread_mutex_destroy(&gps_reader_data.gps_cache_mutex);
+}
+
+
```
pelwell: The other example of extra whitespace.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+{
+   memset(&gps_reader_data, 0, sizeof(gps_reader_data));
+
+   pthread_mutex_init(&gps_reader_data.gps_cache_mutex, NULL);
+
+   gpsd_init(&gps_reader_data.gpsd);
+   if (libgps_load(&gps_reader_data.gpsd))
+   {
+      pthread_mutex_destroy(&gps_reader_data.gps_cache_mutex);
+      exit(EX_SOFTWARE);
```
6by9: exit here seems a tad extreme. Return error and exit from the caller (if that is deemed necessary). Ditto the other failure cases in this function.

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -542,13 +544,13 @@ static void application_help_message(char *app_name)
       fprintf(stdout, ",%s", raw_output_fmt_map[i].mode);
    }
 
-   fprintf(stdout, "\n");
+   fprintf(stdout, "\n\n");
```
6by9: Unrelated change ;-)

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -1139,14 +1140,72 @@ static void update_annotation_data(RASPIVID_STATE *state)
...
+                  (!isnan(gpsdata->fix.altitude)))
+            {
+               snprintf(alt, sizeof(alt), "Altitude=%.2fm",
+                        (gpsdata->fix.altitude*10+0.5)/10.0 );
+            }
+
+            if ((gpsdata->set & SPEED_SET) && !isnan(gpsdata->fix.speed))
+            {
+               snprintf(speed, sizeof(speed), "Speed=%.2fms",
+                        ((gpsdata->fix.speed*MPS_TO_KPH*10+0.5)/10) );
```
6by9: What units are we using where?
Text says milliseconds(!!!), and we're taking a speed (gps.h says in m/s), multiplying by MPS_TO_KPH, and rounding. So the text should say km/h?

host_applications/linux/apps/raspicam/CMakeLists.txt
```diff
@@ -43,9 +43,11 @@ set (COMMON_SOURCES
    RaspiCLI.c
    RaspiPreview.c
    RaspiCommonSettings.c
-   RaspiHelpers.c)
+   RaspiHelpers.c
+   RaspiGPS.c
+   libgps_loader.c)
```
6by9: You've added it to COMMON_SOURCES, so raspividyuv and raspiyuv will now accept the parameter but do nothing with them.

JamesH65: Yes, I thought the benfits of making it common for two of the apps outweighed the fact its not (yet) used in the YUV applications.

host_applications/linux/apps/raspicam/RaspiVid.c
JamesH65: Hmm. Need to double check that. I thoughts it was m/s, but will cross check against the cgps test app.

host_applications/linux/apps/raspicam/RaspiVid.c
JamesH65: Yep....noticed a formating problem while testing. I coudl send it in a different PR, but I have a horrible feeling the life is too short.

host_applications/linux/apps/raspicam/RaspiGPS.c
JamesH65: Hmm, the original dropped out, not nececessarily so abruptly. Will ponder how to make this neater.

host_applications/linux/apps/raspicam/RaspiGPS.c
```diff
@@ -0,0 +1,188 @@
...
+   pthread_mutex_lock(&gps_reader_data.gps_cache_mutex);
+   return &gps_reader_data.gpsdata_cache;
+}
+
+void raspi_gps_unlock()
+{
+   pthread_mutex_unlock(&gps_reader_data.gps_cache_mutex);
+}
+
+
```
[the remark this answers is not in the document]
JamesH65: I really need to adjust my `astyle` parameters to fix that automatically.

host_applications/linux/apps/raspicam/RaspiGPS.c
6by9: Cleanup (not needed here as gpsd_init only does a memset and writes a couple of values) and return an error code.
The caller can abort on error, but it makes this a more useful library than having it randomly abort the app.

The code looks to be doing all the cleanup already, so simply replacing the ```exit``` calls with return -1, and return 0 at the end would do it. (Function needs to change from returning void to returning int too).

host_applications/linux/apps/raspicam/RaspiVid.c
6by9: :-)

host_applications/linux/apps/raspicam/RaspiVid.c
6by9: gps.h says 
```    double speed;	/* Speed over ground, meters/sec */```
so I think it wants to be either 
```"Speed=%.2fkm/h", ((gpsdata->fix.speed*MPS_TO_KPH*10+0.5)/10) ```
or
```"Speed=%.2fm/s", gpsdata->fix.speed) ```

host_applications/linux/apps/raspicam/CMakeLists.txt
6by9: I'll let you off then.

JamesH65: Phew

host_applications/linux/apps/raspicam/RaspiVid.c
JamesH65: Yup, a number of those calcs can be much simplified. Another C&P thingy.

JamesH65:
OK, so this latest push is a little more than just the cleanups suggested above. I have moved the location string generation to the RaspGPS file so that it can now be used in all the apps and have modified all the apps to use it. So you can now get the GPS annotation on stills and video, in normal or yuv/rgb form.

JamesH65:
Thanks all. Except for the extra line at the end comment. Grr.

PR #519, opened 2018-12-19
Added H264 slicing.

JamesH65:
This is a modified version of that provided by shizukachan

We are having to limit this to widths < 1280 due to an as yet
unfixed firmware issue. I've taken a quick look - its not
obvious what is going on.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -224,6 +224,7 @@ struct RASPIVID_STATE_S
 
    bool netListen;
    MMAL_BOOL_T addSPSTiming;
+   int slices;
 };
 
 
@@ -306,7 +307,8 @@ enum
    CommandRaw,
    CommandRawFormat,
    CommandNetListen,
-   CommandSPSTimings
+   CommandSPSTimings,
+   CommandSlices
 };
 
 static COMMAND_LIST cmdline_commands[] =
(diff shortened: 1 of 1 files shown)


host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2020,8 +2034,29 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
...
    }
 
+   if (state->encoding == MMAL_ENCODING_H264 && state->slices > 1 && state->common_settings.width <= 1280)
+   {
+      int frame_mb_rows = VCOS_ALIGN_UP(state->common_settings.height, 16) >> 4;
+
+      if (state->slices > frame_mb_rows) //warn user if too many slices selected
+      {
+         fprintf(stderr,"H264 Slice count (%d) exceeds number of macroblock rows (%d). Setting slices to %d.\n", state->slices, frame_mb_rows, frame_mb_rows);
+         // Continue rather than abort..
```
6by9: Does it actually continue in this case, or do we need to revert to 1 slice?

host_applications/linux/apps/raspicam/RaspiVid.c
```diff
@@ -2020,8 +2034,29 @@ static MMAL_STATUS_T create_encoder_component(RASPIVID_STATE *state)
...
+      }
+      int slice_row_mb = frame_mb_rows/state->slices;
+      if (frame_mb_rows - state->slices*slice_row_mb)
+         slice_row_mb++; //must round up to avoid extra slice if not evenly divided
```
6by9: Answering my own question, I suspect it does as ```slice_row_mb``` will become 0, ```frame_mb_rows - 0``` is true, so we end up with ```slice_mb_row``` as 1 due to the ++.
It'd be nice to confirm this though.

JamesH65: @shizukachan Can you comment on this?

6by9:
Just noticed that #486 has been merged which looks nearly identical to this. How do the two interact?

JamesH65:
Odd, I did a search and couldnt find this branch being merged, so PR'd it, but its looks like its already there anyway. I'll cancel this one since I think its all there anyway.

PR #520, opened 2019-01-07
Move to standard Linux  microsecond count

JamesH65:
The camera apps were using vcos_getmicrosecs64 which
was in turn using gettimeofday. This means timelapse code
goes very wrong when the date/time is changed while the
timelapse is being done.

Moved to a scheme based on clock_gettime(CLOCK_MONOTONIC_RAW)
which is not affected by underlying clock changes.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiHelpers.c
+++ b/host_applications/linux/apps/raspicam/RaspiHelpers.c
@@ -280,3 +280,17 @@ int mmal_status_to_int(MMAL_STATUS_T status)
       return 1;
    }
 }
+
+
+uint64_t get_microseconds64()
+{
+   struct timespec spec;
+   uint64_t us;
+
+   clock_gettime(CLOCK_MONOTONIC_RAW, &spec);
+
+   us = spec.tv_sec * 1000000;
+   us += spec.tv_nsec / 1000;
+
+   return us;
+}
--- a/host_applications/linux/apps/raspicam/RaspiHelpers.h
+++ b/host_applications/linux/apps/raspicam/RaspiHelpers.h
@@ -38,5 +38,6 @@ void check_disable_port(MMAL_PORT_T *port);
 void default_signal_handler(int signal_number);
 int mmal_status_to_int(MMAL_STATUS_T status);
 void print_app_details(FILE *fd);
+uint64_t get_microseconds64();
 
 #endif
--- a/host_applications/linux/apps/raspicam/RaspiStill.c
+++ b/host_applications/linux/apps/raspicam/RaspiStill.c
@@ -1427,7 +1427,7 @@ static int wait_for_next_frame(RASPISTILL_STATE *state, int *frame)
    static int64_t complete_time = -1;
    int keep_running = 1;
 
-   int64_t current_time =  vcos_getmicrosecs64()/1000;
+   int64_t current_time =  get_microseconds64()/1000;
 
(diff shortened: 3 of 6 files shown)


host_applications/linux/apps/raspicam/RaspiHelpers.c
```diff
@@ -280,3 +280,17 @@ int mmal_status_to_int(MMAL_STATUS_T status)
       return 1;
    }
 }
+
+
+uint64_t get_microseconds64()
+{
+   struct timespec spec;
+   uint64_t ms;
```
popcornmix: As ``ms`` usually means milliseconds, this made the following code look very wrong to me.
I'd prefer ``us`` or ``microsecs`` or ``microseconds`` to avoid the confusion.

JamesH65: Absolutely right - a C&P from code that worked in milliseconds. Will fix up tomorrow.

JamesH65:
Have changed ms to us and pushed.

popcornmix:
This PR is fine, but I wonder if changing the implementation of ``vcos_getmicrosecs64`` to match this would be even better.
I suspect ``vcos_getmicrosecs64`` is only ever used for time deltas, and the issues here (NTP or user adjusting clock) could probably affect other callers.

popcornmix:
Okay, we'll stick with the lower risk option.

lurch:
Presumably there wouldn't be much point in adding a new `vcos_getmicrosecs64_monotonic` ?

lurch:
Looking at the diff in this PR, it seems `get_microseconds64` is _only_ ever called as `get_microseconds64()/1000;` or `get_microseconds64()/1000000;` so I guess you _could_ have simply implemented `get_milliseconds64` instead? :wink:

6by9:
> Presumably there wouldn't be much point in adding a new ```vcos_getmicrosecs64_monotonic``` ?

Not really. ```vcos_getmicrosecs64``` should be that anyway, but was implemented in a bizarre manner by someone back in the very old Broadcom days (the implementation hasn't changed since the branch was taken in 2012!)
I also don't think there should be any repercussions from correcting the implementation, but it would need a short discussion to try and think up any obscure uses it could be currently put to.
The function description (interface/vcos/vcos_thread.h) is
```
/**
  * \brief Return the value of the hardware microsecond counter.
  */
```
so that would imply that we want to switch to CLOCK_MONOTONIC_RAW.

lurch:
Is it theoretically possible that somebody might be abusing `vcos_getmicrosecs64` to calculate the time-of-day? :man_shrugging: 

AIUI (which could be wrong!) the `vcos_` stuff all calls into the VPU/GPU-side of things, but CLOCK_MONOTONIC_RAW is a Linux-side thing?

6by9:
> Is it theoretically possible that somebody might be abusing vcos_getmicrosecs64 to calculate the time-of-day?

Hopefully not, but it's that sort of thing that needs to be discussed.

> AIUI (which could be wrong!) the vcos_ stuff all calls into the VPU/GPU-side of things,

No, ```vcos_*``` are OS abstraction functions to allow code to run under multiple OSes. There are implementations for pthreads (Linux), threadx (VPU), Symbian, Nucleus, and win32.
How the implementation achieves the desired functionality is entirely down to the implementer, and in this case they've done a dubious job of it.

PR #525, opened 2019-01-14
fix vcfiled cross-compile install

jwinarske:
Cross compile install fails with vcfiled, as it is attempting to write /etc/init.d.  This should be ${CMAKE_INSTALL_PREFIX}/etc/init.d.

### DIFF
--- a/interface/vmcs_host/linux/vcfiled/CMakeLists.txt
+++ b/interface/vmcs_host/linux/vcfiled/CMakeLists.txt
@@ -21,7 +21,7 @@ configure_file (etc/init.d/vcfiled ${PROJECT_BINARY_DIR}/etc/init.d/vcfiled)
 
 # script to start up vcfiled at start of day
 install(PROGRAMS ${PROJECT_BINARY_DIR}/etc/init.d/vcfiled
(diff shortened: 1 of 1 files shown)


popcornmix:
Possibly dropping vcfiled completely would be good (it's not used), but your change looks like an improvement.

popcornmix:
@6by9 no objections to deletion (it won't be coming back).
Technically if it's not built from vc4 tree it will result in deletion from userland tree (only built source is included).

6by9:
> Technically if it's not built from vc4 tree it will result in deletion from userland tree (only built source is included).

If I've understood you correctly then it should automagically purge itself once you've picked this change to vc4 and rerun your scripts? In which case I won't create userland PRs to remove it.

popcornmix:
> If I've understood you correctly then it should automagically purge itself once you've picked this change to vc4 and rerun your scripts?

This PR doesn't stop vcfiled from being built though.

6by9:
> This PR doesn't stop vcfiled from being built though.

:-) Nor it does.
I'll create a vc4 PR to drop it from the build, so hopefully that will ripple through easily.

PR #528, opened 2019-01-15
vcgencmd: add basic usage, detect when Videcore returns an error

andrum99:
Prompted by https://github.com/raspberrypi/documentation/issues/1065 and my general annoyance with vcgencmd having no built-in help whatsoever, I've tweaked it to return basic usage when:

- it is called with no parameters
- it is called with -? as the first parameter
- it is called with --help as the first parameter

In addition, it now also sets its return value to indicate an error if the Videocore returned an error.

And finally, if the error Videocore returned was that the command was not recognised, suggest 'vcgencmd commands' to get a list of possible commands.


andrum99:
This should be seen as a starting point for the documentation. The next step, I would suggest, is to add a man page.

JamesH65:
Thanks! Out of interest, how does the man file make it to the correct place when things are compiled/installed?

andrum99:
At the moment it doesn't. Do feel free to add to it 😉

pelwell:
"VideoCore" should be spelled like that everywhere that it appears.
Adding yourself as a Copyright holder is a bit of a stretch, don't you think?

Otherwise it looks OK.

andrum99:
In what way is it a stretch? I wrote the code I've proposed adding.

pelwell:
The Linux kernel sources usually list the copyright of the primary author(s), and the author(s) of works from which it was derived. Secondary contributors are recorded in commit messages - unless you can provide me with a counter-example.

andrum99:
I'm happy to go with that convention, which I was not aware of, noting of
course that this is not part of the Linux kernel.

pelwell:
Looks good to me.

andrum99:
I will leave actually building the man page to a follow up PR. Let me know if you would prefer the man page source excised from this one.

Also, are there any other consumers of the VideoCore command interface this uses? I'm wondering if it would best be documented on its own separate man page.

JamesH65:
LGTM

andrum99:
Please add my name as Andrew Pattison with email address andrum99@gmail.com when you commit this to the master repository, or advise how I should add it myself at this stage.

JamesH65:
Usual way to add a Signed Off tag in the commit message. 

Here's one from a commit message in the the Linux kernel

Signed-off-by: James Hughes <james.hughes@raspberrypi.org>

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -71,6 +86,14 @@ int main( int argc, char **argv )
 
     if (argc > 1)
     {
+       // first check if we were invoked with either -? or --help
+       // in which case show basic usage and exit
+      if( ( strcmp( argv[1], "-?" ) == 0) || strcmp( argv[1], "--help" ) == 0 )
```
lurch: I've definitely seen `--help` used before, but is `-?` a common flag for Linux CLI programs?

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -130,6 +153,14 @@ int main( int argc, char **argv )
           {
               printf("%s\n", buffer );
           }
+          if (strncmp( buffer, "error=", 5) == 0 )
+          {
+             if ( strcmp( buffer, "error=1 error_msg=\"Command not registered\"" ) == 0 )
+             {
+                printf( "Use 'vcgencmd commands' to get a list of commands\n" );
```
lurch: Should this be printing to STDERR?

host_applications/linux/apps/gencmd/gencmd.1
```diff
@@ -0,0 +1,24 @@
+.TH vcgencmd 1 "January 2019" "Raspberry Pi vcgencmd" "User Commands"
+.SH NAME
+vcgencmd - send a command to the Videcore
```
lurch: VideoCore here?

andrum99: fixed

host_applications/linux/apps/gencmd/gencmd.c
andrum99: Well I've seen it used and it has 2 LGTMs from reviewers so I'm happy to leave it in unless @popcornmix objects

host_applications/linux/apps/gencmd/gencmd.c
andrum99: Good point. None of the other errors are getting printed to STDERR so I think I will leave that to a future PR.

host_applications/linux/apps/gencmd/gencmd.c
pelwell: -h is more usual.

andrum99:
Thanks - I've squashed it down and minified the commit message (second force-push) so it should be ready to roll, once it has the necessary approvals.

host_applications/linux/apps/gencmd/gencmd.c
andrum99: I will change it to -h in a future PR, since I'm going to be creating one to have it output errors to STDERR anyway and I've already squashed my commits down. Unless anyone has any objections?

JamesH65:
You could use `git commit --amend` for new staged changes so they are just added to the current commit rather than creating a new one, that's what i ussually do. It's probably best to do all the changes in one go.

pelwell:
Provided the end result is a single commit and the follow-ups are essentially bug fixes, it can be useful to keep them separate and squash when merging because it makes reviewing easier.

JamesH65:
Go for it!

popcornmix:
No objections here (and I agree ``-h`` is preferable to ``-?``).

host_applications/linux/apps/gencmd/gencmd.c
andrum99: I've now got all errors going to stderr

host_applications/linux/apps/gencmd/gencmd.c
andrum99: Changed -? to -h

andrum99:
Okay, I think that's all the review comments dealt with now. I've removed the man page from this PR as I need to go and investigate how to actually get this built. If could get one final review, then I can squash it down into a single commit (again).

pelwell:
Don't worry about the final squash - we can do that at merge time (as long as the desired result is a single commit).

andrum99:
I've fixed the wording of how vcgencmd reports how long the command took to return. But there is some rather odd code where it returns the time taken in seconds, millieseconds and microseconds simply by multiplying by 1 000 and 1 000 000. Can I just delete that as it seems to serve no useful purpose? Here is the code:

`printf( "Time taken: %f seconds (%f msecs) (%f usecs)\n", time_diff, time_diff * 1000, time_diff * 1000000 );`

lurch:
I suspect the time-taken stuff was probably added purely for debug purposes. And when doing this kind of debugging, it's much easier to at-a-glance see the difference between 221 usecs and 21 usecs than it is to tell the difference between 0.000221 seconds and 0.000021 seconds :wink:

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -36,6 +36,18 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "interface/vmcs_host/vc_vchi_gencmd.h"
 
+void show_usage()
+{
+   printf( "Usage: vcgencmd [-t] [COMMAND]\n" );
```
lurch: Only a very minor point, but I wonder if `vcgencmd [-t] [commands|COMMAND]` might be better here? :man_shrugging:

andrum99:
OK folks, so this is now ready to go, except that it is dependent on @JamesH65 vcgencmd documentation landing in the documentation repo. The reason being it now has the proposed URL of the page in the docs as part of the usage. Can I get some reviews of the C code if you're not too busy?

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -36,6 +36,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+   printf( "Usage: vcgencmd [-t] command\n" );
+   printf( "Send a command to the VideoCore and print the result.\n\n" );
+   printf( "  -t         Time how long the command takes to complete\n" );
+   printf( "Use the command `vcgencmd commands` to get a list of available commands\n\n" );
+   printf( "Exit status:\n" );
+   printf( "   0    command completed successfully\n" );
+   printf( "  -1    problem with VCHI\n" );
+   printf( "  -2    VideoCore returned an error\n\n" );
+   printf( "For further documentation please see\n" );
+   printf( "raspberrypi.org/documentation/raspbian/applications/vcgencmd.md\n\n" );
```
lurch: Is it common practice to put URLs in help strings? :man_shrugging: 
In any case, it should probably be `https://www.rasperrypi.org/...` ?

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -36,6 +36,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "interface/vmcs_host/vc_vchi_gencmd.h"
 
+void show_usage()
+{
+   printf( "Usage: vcgencmd [-t] command\n" );
```
lurch: @pelwell / @popcornmix What's the `instNum` for? Should that be documented in the usage string too?

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -50,27 +64,38 @@ int main( int argc, char **argv )
...
         return -1;
     }
 
     vc_vchi_gencmd_init(vchi_instance, &vchi_connection, 1 );
 
     if (argc > 1)
     {
+       // first check if we were invoked with either -h or --help
+       // in which case show basic usage and exit
+      if( ( strcmp( argv[1], "-h" ) == 0) || strcmp( argv[1], "--help" ) == 0 )
```
lurch: Minor inconsistency in that the condition to the left of the `||` has an extra pair of brackets than the condition to the right of the `||`.

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -128,7 +153,19 @@ int main( int argc, char **argv )
           }
           else
           {
-              printf("%s\n", buffer );
+             if (strncmp( buffer, "error=", 5) == 0 )
```
lurch: "error=" is 6 characters long, not 5 :wink:

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -36,6 +36,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "interface/vmcs_host/vc_vchi_gencmd.h"
 
+void show_usage()
+{
+   printf( "Usage: vcgencmd [-t] command\n" );
+   printf( "Send a command to the VideoCore and print the result.\n\n" );
+   printf( "  -t         Time how long the command takes to complete\n" );
+   printf( "Use the command `vcgencmd commands` to get a list of available commands\n\n" );
```
lurch: Would single-quotes be more appropriate here than backticks? :man_shrugging:

host_applications/linux/apps/gencmd/gencmd.c
pelwell: Absolutely nothing at all - it's a function-local variable which is assigned to and never used.

lurch: Cool, that's what I thought, but as this is code which talks to firmware I didn't know if there was some deeper magic going on :wink:

host_applications/linux/apps/gencmd/gencmd.c
andrum99: That was based what vcdbg outputs when it prints its usage. I'm open to suggestion as to which one is correct.

host_applications/linux/apps/gencmd/gencmd.c
andrum99: No idea. Since there is no man page, and the command is fully documented on the website, I've added a link to it. I don't really care if it is not common practice.

andrum99:
Thanks for catching those @lurch 😀 I've changed everything per your suggestions.

host_applications/linux/apps/gencmd/gencmd.c
```diff
@@ -41,30 +41,22 @@ void show_usage()
...
    printf( "  -t         Time how long the command takes to complete\n" );
-   printf( "Use the command `vcgencmd commands` to get a list of available commands\n\n" );
+   printf( "Use the command 'vcgencmd commands' to get a list of available commands\n\n" );
    printf( "Exit status:\n" );
    printf( "   0    command completed successfully\n" );
    printf( "  -1    problem with VCHI\n" );
    printf( "  -2    VideoCore returned an error\n\n" );
    printf( "For further documentation please see\n" );
-   printf( "raspberrypi.org/documentation/raspbian/applications/vcgencmd.md\n\n" );
+   printf( "https://raspberrypi.org/documentation/raspbian/applications/vcgencmd.md\n\n" );
```
lurch: Can you add the `www` as well?

andrum99:
Please close this orphaned PR and add @pelwell's approval to the new PR. (It's a copy and paste job of the whole file from this PR).

lurch:
@JamesH65 This should be closed - it was replaced with #552

JamesH65:
As per, closing.

PR #530, opened 2019-02-21
raspivid: Stop --raw option resetting --raw-format to YUV

6by9:
Observed on
https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=189830&start=25#p1432589

The raspivid command line
raspivid --raw-format rgb --raw foo.rgb -o foo.h264
actually saved YUV data to foo.rgb as the --raw option reset
raw-format to YUV mode.

Renumber the enums so that YUV is 0, and use the fact the state
structure is memset to 0 (and therefore YUV) to avoid the
-raw parser needing to change the format at all.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -158,7 +158,7 @@ typedef struct
  */
 typedef enum
 {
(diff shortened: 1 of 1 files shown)


6by9:
@JamesH65 Trivial one for you to review.

JamesH65:
LGTM

PR #535, opened 2019-03-04
Fixed up overflow in microsecond timer

JamesH65:
Was using a 32 bit calculation when should have been 64bit.


JamesH65:
Note: deliberately left some not quite right mismatches in types (assigning uint to int etc), they will be fixed in a future PR. This just fixes the issue in hand, of timelapse mode dying after an hour or so due to a 32 bit overflow.

PR #536, opened 2019-03-04
2 makefiles

tvjon:
@6by9 
Sorry for the errors, & thank you for tidying it up.

BTW, I have simple omx-il  PCM-file-to-render audio (as opposed to the sine wave table in hello_audio) working, & am wondering if I can somehow "wrap" it to test in mmalplay? I'd need to mux a h264  with pcm audio in a suitable container so would make for a biggish file, but at least it'd let me see a path for audio to work in mmalplay. I realise you're busy, but any pointers would be very helpful. Oddly my omx-il version doesn't have the irritating buzz that's present in the hello_audio.bin example. I don't remember the buzz being there a couple of years ago, strange.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,4 @@ EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 Use buildme to build. It requires cmake to be installed and an arm cross compiler. It is set up to use this one:
 https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian
 
-Note that this repository does not contain the source for the edid_parser and vcdbg binaries due to licensing restrictions.
+Note that this repository does not contain the source for the edidparser and vcdbg binaries due to licensing restrictions.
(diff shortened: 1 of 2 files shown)


host_applications/vmcs/test_apps/mmalcam/Makefile
```diff
@@ -0,0 +1,2 @@
+all: 
+	gcc -O3  viewfinder.c mmalcam.c -o mmalplay -lmmal -lmmal_core  -lmmal_util  -I/opt/vc/include/  -L/opt/vc/lib/ -lpthread -lvcos
```
6by9: Minor objection to ```-o mmalplay```!
I'll fix this up in the cherry-pick.

6by9:
Closing as superceded by #537 just for tidying up.

Please can we have the discussion over your audio tweaks via either the forum, an issue (https://github.com/raspberrypi/userland/issues), or push the PR and we'll review it.
The golden rule of issue trackers is to always keep it to one issue/discussion per thread, otherwise life gets confusing.

Having large binaries in source control is generally a bad plan, but a few second snippet of audio at lowish sample rate wouldn't be too huge.

PR #537, opened 2019-03-04
Makefiles for mmalcam and mmalplay

6by9:
Supercedes #536 as @tvjon is finding his way with git/github.

Minor update over the original in the output file for mmalcam has been amended to be ```mmalcam``` instead of ```mmalplay```


JamesH65:
LGTM

PR #538, opened 2019-03-04
buildme: Add option for ARM64 builds via either native or cross compile

6by9:
Supports native builds on aarch64 machines.

Supports making a 64bit build via either cross compiling, or on
aarch64 machines.
Both options require the addition of --arm64 to the command line,
otherwise 32bit libraries will be built.

Cross compiling requires the aarch64-linux-gnu- compile tools to
be installed (packages gcc-5-aarch64-linux-gnu and g++-aarch64-linux-gnu
on Ubuntu).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. It is set up to use this one:
+Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32 bit cross compiliation it is set up to use this one:
 https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian
+
+Whilst 64 bit userspace is not officially supported, some of the libraries will work for it. To cross compile, install gcc-aarch64-linux-gnu and g++-aarch64-linux-gnu first. For both native and cross compiles, add the option ```--aarch64``` to the buildme command.
 
 Note that this repository does not contain the source for the edidparser and vcdbg binaries due to licensing restrictions.
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32 bit cross compiliation it is set up to use this one:
+Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32-bit cross compiliation it is set up to use this one:
 https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian
 
-Whilst 64 bit userspace is not officially supported, some of the libraries will work for it. To cross compile, install gcc-aarch64-linux-gnu and g++-aarch64-linux-gnu first. For both native and cross compiles, add the option ```--aarch64``` to the buildme command.
+Whilst 64-bit userspace is not officially supported, some of the libraries will work for it. To cross compile, install gcc-aarch64-linux-gnu and g++-aarch64-linux-gnu first. For both native and cross compiles, add the option ```--aarch64``` to the buildme command.
 
 Note that this repository does not contain the source for the edidparser and vcdbg binaries due to licensing restrictions.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32-bit cross compiliation it is set up to use this one:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,7 +2,9 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. It is set up to use this one:
+Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32 bit cross compiliation it is set up to use this one:
 https://github.com/raspberrypi/tools/tree/master/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian
 
+Whilst 64 bit userspace is not officially supported, some of the libraries will work for it. To cross compile, install gcc-aarch64-linux-gnu and g++-aarch64-linux-gnu first. For both native and cross compiles, add the option ```--aarch64``` to the buildme command.
```
lurch: It says `--aarch64` here, but `--arm64` in the buildme ?

README.md
```diff
@@ -2,7 +2,9 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. It is set up to use this one:
+Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32 bit cross compiliation it is set up to use this one:
```
pelwell: I like to hyphenate `<n>-bit`.

6by9:
Thanks both @lurch and @pelwell.

Amendments made generally via the fixups patch, but also in the commit text of the first one. I have standardised on --aarch64 which wasn't what the original commit text had.

README.md
```diff
@@ -2,7 +2,9 @@ This repository contains the source code for the ARM side libraries used on Rasp
 These typically are installed in /opt/vc/lib and includes source for the ARM side code to interface to:
 EGL, mmal, GLESv2, vcos, openmaxil, vchiq_arm, bcm_host, WFC, OpenVG.
 
-Use buildme to build. It requires cmake to be installed and an arm cross compiler. It is set up to use this one:
+Use buildme to build. It requires cmake to be installed and an arm cross compiler. For 32-bit cross compiliation it is set up to use this one:
```
JamesH65: speeling compiliation

buildme
```diff
@@ -1,19 +1,28 @@
...
 	BUILDTYPE=Debug
 	shift
 fi
 
+if [ "$1" = "--aarch64" ]; then
+	ARM64=ON
+	CMAKE_TOOLCHAIN_FILE=../../../makefiles/cmake/toolchains/aarch64-linux-gnu.cmake
+	shift
+fi
+
```
JamesH65: No ability to have a debug 64 bit build?

README.md
JamesH65: Should ARM be capitalised in some way?

6by9: OK

buildme
6by9: Yes you can. ```./buildme --debug --aarch64```.
```shift``` removes the first command line argument and shifts the others up.

It is true that ```./buildme --aarch64 --debug``` will give you a 64-bit release build due to the parsing order. I'm not a bash expert so not intending to fix that one.

Comment (pelwell) [reply buildme]: Agreed - that a debug build is possible without editing the script is enough.

**Comment by 6by9** [comment]
FWIW I have tried this on a Gentoo64 image and it seems to build a working vcgencmd and others. ```file``` identifies a locally built binary as:
```
demouser@pi64 ~ $ file /opt/vc/bin/vcgencmd 
/opt/vc/bin/vcgencmd: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, not stripped
```
so that would appear to be the build that we want.
Cross compiled I get
```
build/bin/vcgencmd: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, BuildID[sha1]=3cb5f5bcde9edbaa84933ea32a8c9e423f306d2b, not stripped
```
Not sure what the difference with ```ELF 64-bit LSB **pie** executable``` means. Where's Google.....

**Comment by pelwell** [comment]
Position Independent Executable?

**Comment by 6by9** [comment]
PIE = Position Independent Executable.
The natively built executable for vcgencmd is 24072 bytes, vs 20496 bytes for the cross-compiled.

**Comment by 6by9** [comment]
So the next question is do we care? Checking if we get the same anomally on 32 bit....

**Comment by pelwell** [comment]
You can either force pie with `--pie` or disable it with `--no-pie`, but I don't think it matters either way - it's just that the toolchains have different defaults.

**Comment by 6by9** [comment]
Both report basically the same on 32-bit, just version numbers change
Native (Buster)
```
/opt/vc/bin/vcgencmd: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for **GNU/Linux 3.2.0**, BuildID[sha1]=6c030e30f53f26614e8d072d8858b0c3618643bd, not stripped
```
```
build/bin/vcgencmd: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for **GNU/Linux 2.6.26**, BuildID[sha1]=b2d1e43261caa4f45bb8dc995850aeaafdd99cff, not stripped
```

lurch:
Raspberry PIE? :cake: :yum:

6by9:
If we're not fussed then I'm not going to fight CMake to get the flags right.

All happy now?

buildme
JamesH65: I had completley missed the shift, sorry.

pelwell: Shift happens.

PR #541, opened 2019-03-14
debug_sym: Remove annoying message should fb copy fail

6by9:
Any kernel not using bcm2708_fb as the FB driver is likely to
fail the FBIODMACOPY ioctl that vcdbg tries to use to access
the hidden 16MB of RAM behind the peripheral space.
Remove the error message as it drops back to using /dev/mem
instead anyway.

Cherry-pick of 21a576f9f74df33f233b18717d225867c912e818 from the firmware tree as that appears not to have happened automagically.

### DIFF
--- a/host_applications/linux/libs/debug_sym/debug_sym.c
+++ b/host_applications/linux/libs/debug_sym/debug_sym.c
@@ -165,8 +165,6 @@ static int vc_mem_copy(void *dst, uint32_t src, uint32_t length)
 
     if ( ioctl( memFd, FBIODMACOPY, &ioparam ) != 0 )
     {
-        ERR( "Failed to get memory size via ioctl: %s(%d)\n",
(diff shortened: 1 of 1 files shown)


pelwell:
I approve, but want to give the magic time to work...

6by9:
I'm not sure how much is fully automagic, and how much is @popcornmix noticing that the patch also affects userland.
This was MR !290.

pelwell:
Don't you think @popcornmix is magic?

6by9:
> Don't you think @popcornmix is magic?

Frequently!

6by9:
The magic hasn't happened in 6 months, so can we merge this?
(I accidentally force pushed the branch as I thought it was in, and then found it wasn't. Branch reset to where it was)

PR #544, opened 2019-03-22
Fix header enum typo

lurch:
Old enum value also retained for backwards-compatibility :slightly_smiling_face: 
(although I can remove it if you don't think it's needed)


Ruffio:
Shouldn't this be merged?

PR #553, opened 2019-05-29
Fix for missing timestamps when data exceed one buffers worth

JamesH65:
Fix provided by DaveS.

The addition of a NAL flag appears to have broken timestamps with
large buffers. This simplifies the 'if' around timestamps and takes
the NAL change in to account.

Also fixed up some bad code formatting after the if statement. Yes, I know.


6by9:
Nit picks.

If fixing the code style, then adding the space after the if in ```if(pData->pstate->save_pts``` would have been nice.

No need to put the comment about where the fix came from in the commit text.

JamesH65:
Fixed up Dave's comments.

PR #554, opened 2019-06-03
Add media flushing call on when writing bitstream

JamesH65:
Under certain circumstances, the Linux file system write behind caching
get so much data to write in one go that is causes frames to be dropped.
The current fflush is not sufficient as that only flushes the C library
caches, not the Linux filesystem caches.

This fix forces data to be written to media as soon as it arrives, using
fdatasync().

Note, currently it's conditional on the --flush  command line parameter.


6by9:
> As long as this is an optional feature I'm happy for fdatasync to be called as is, but perhaps a file size increment threshold could be applied to reduce the number of writes if it becomes the default.

Agreed that would needed before this is default. Trigger after every N MBytes as otherwise we're also increasing the SD card write operations significantly.

PR #555, opened 2019-06-04
Fix for missing cleanup step for hello_videocube and hello_teapot

paolosabatino:
The cleanup step in hello_videocube and hello_teapot is never executed because some bits are missing.
This pull request adds intercepts SIGINT and issue the cleanup when the user pressed CTRL-C.
Also this fixes the egl_render freeze when the OMX_StateLoaded is requested and properly disposes the buffer obtained with OMX_UseEGLImage() call

### DIFF
--- a/host_applications/linux/apps/hello_pi/hello_teapot/triangle.c
+++ b/host_applications/linux/apps/hello_pi/hello_teapot/triangle.c
@@ -46,6 +46,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "triangle.h"
 #include <pthread.h>
 
+#include <signal.h>
+
 
 #define PATH "./"
 
@@ -94,6 +96,8 @@ static CUBE_STATE_T _state, *state=&_state;
 
 static void* eglImage = 0;
 static pthread_t thread1;
+
+extern int thread_run;
 
 
 /***********************************************************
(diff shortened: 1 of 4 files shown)


JamesH65:
Coding style of PR isn't matching the code - please use Allman with 3 spaces instead of tabs.  Cannot comment on the technical aspects! @6by9?

6by9:
See discussion on https://www.raspberrypi.org/forums/viewtopic.php?f=70&t=241923

6by9:
As per the thread on the forums, the more consistent approach would be to disable egl_render's output port (with the special handling for manually calling OMX_FreeBuffer on the buffers), and then the ilclient_state_transition calls should run without issue.

6by9:
I'd have preferred the the egl_render output port disable to be with all the other ones before ```ilclient_state_transition(list, OMX_StateIdle);```, but I'm not that fussed.

paolosabatino:
There's a reason: if the port is disabled and the buffer freed after the transition to Idle, the fill buffer callback can check against component state and avoids a spurious call to OMX_FillThisBuffer() and an error report

Disabling the port and freeing the buffer is not sufficient to avoid the callback from the omx component

6by9:
OK, that's reasonable, although there may be a potential for deadlock. Calling back into IL from a callback context is generally advised against.
Most of the example code will put callback buffers onto a queue for a main thread to pick up and do something with. The callback context work is kept to an absolute minimum.

I'm happy. Anyone else want to review before squash and merge?

popcornmix:
Ok by me.

6by9:
Sorry for the delay - now merged.
Thanks for your contribution.

PR #561, opened 2019-07-17
vcgencmd: Increase buffer size and ensure strings are null terminated

timg236:
The current maximum response size from the firmware is 4096. Increase
the client's buffer to match this.

Additionally, ensure that the buffer is null terminated, otherwise,
strlen could fail.


timg236:
@popcornmix @pelwell

popcornmix:
Might also need to increase ``GENCMDSERVICE_MSGFIFO_SIZE``

timg236:
> Might also need to increase `GENCMDSERVICE_MSGFIFO_SIZE`

Ok, increasing this to 4095 which is the maximum size supported in a single VCHI cmd

PR #567, opened 2019-08-05
vcsm: Add support for vcsm_vc_addr_from_hdl to vcsm-cma path

6by9:
It does have potential nasty failure mechanisms should you be
using 64 bit kernels etc, but should work for all current cases.


popcornmix:
Okay with me.

6by9:
Any further comments on this, or can I merge?
@JamesH65

JamesH65:
LGTM

PR #572, opened 2019-09-03
raspicam: Add -dn option to set the display number

6by9:
Allows options like -dn 7 to push the preview onto HDMI1.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiPreview.c
+++ b/host_applications/linux/apps/raspicam/RaspiPreview.c
@@ -49,7 +49,8 @@ enum
    CommandPreview,
    CommandFullScreen,
    CommandOpacity,
-   CommandDisablePreview
+   CommandDisablePreview,
+   CommandDisplayNum,
 };
 
 static COMMAND_LIST cmdline_commands[] =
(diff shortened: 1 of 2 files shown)


pelwell:
How does `-dn 7` result in output on HDMI1? What are the legal display numbers?

6by9:
Takes the standard Dispmanx display enums - https://github.com/raspberrypi/userland/blob/master/interface/vmcs_host/vc_dispmanx_types.h#L58

pelwell:
I predict confusion. At the very least the usage info should say that, but perhaps a string would be more user friendly.

popcornmix:
We do use the same numbering in other places (tvservice, omxplayer)
Although tvservice does help you work it out.
```
pi@b3plus:~ $ tvservice --help
Usage: tvservice [OPTION]...
...
  -l, --list                        List all attached devices
  -v, --device                      Specify the device to use (see --list)
pi@b3plus:~ $ tvservice --list
1 attached device(s), display ID's are : 
Display Number 2, type HDMI 0
Display Number 7, type HDMI 1
```
Not sure if a simple string could explain it.

6by9:
Perhaps a string is nicer, although .

It's the same numbers as have been previously used with omxplayer
https://www.raspberrypi.org/documentation/raspbian/applications/omxplayer.md
> If you are using the Raspberry Pi Foundation's touchscreen display, and you want to use it for video output, use the display option to specify which display to use. n is 5 for HDMI, 4 for the touchscreen. With the Raspberry Pi 4 you have two options for HDMI output. n is 2 for HDMI0 and 7 for HDMI1.
>
> omxplayer --display n /opt/vc/src/hello_pi/hello_video/test.h264

Dispmanx display numbers is a mess anyway. The numbers get remapped if the display specified doesn't exist, so 0 (nominally MAIN LCD) will go to HDMI0 if neither a DSI or DPI display is connected.

pelwell:
I'd settle for a reference to `dispmanx numbering` or `tvservice --list`.

6by9:
Help text string and commit text updated.

pelwell:
Thanks - looks good to me.

PR #576, opened 2019-09-27
bcm_host and vcsm patches

6by9:
Cherry-picking @popcornmix 's patch from the firmware branch, and adding one for vcsm-cma cache flushing.

### DIFF
--- a/host_applications/linux/libs/bcm_host/bcm_host.c
+++ b/host_applications/linux/libs/bcm_host/bcm_host.c
@@ -26,6 +26,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 #include <stdio.h>
+#include "include/bcm_host.h"
 #include "interface/vmcs_host/vc_dispmanx.h"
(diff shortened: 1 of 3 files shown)


host_applications/linux/libs/bcm_host/bcm_host.c
```diff
@@ -168,4 +169,148 @@ unsigned bcm_host_get_sdram_address(void)
...
+}
+
+
+static int bcm_host_is_fkms_or_kms_active(int kms)
+{
+   if (!read_string_from_file("/proc/device-tree/soc/v3d@7ec00000/status", "okay", NULL) &&
+       !read_string_from_file("/proc/device-tree/v3dbus/v3d@7ec04000/status", "okay", NULL))
+      return 0;
+   else
+      return read_string_from_file("/proc/device-tree/soc/firmwarekms@7e600000/status", "okay", NULL) ^ kms;
```
pelwell: It seems a strange implementation - looking for a particular mode rather than asking what the mode is - but it looks like it should work.

pelwell:
Looks OK to me.

PR #578, opened 2019-10-03
vcsm: If /dev/vcsm fails, then try /dev/vcsm-cma not already tried.

6by9:
If export is not requested, then /dev/vcsm was the only device tried.

On aarch64 we are not including /dev/vcsm as it has too many
32 bit assumptions, therefore the sensible thing is to try
/dev/vcsm-cma, even if export hasn't been explicitly requested.

### DIFF
--- a/host_applications/linux/libs/sm/user-vcsm.c
+++ b/host_applications/linux/libs/sm/user-vcsm.c
@@ -273,13 +273,22 @@ int vcsm_init_ex( int want_export, int fd )
       goto out; /* VCSM already opened. Nothing to do. */
    }
 
+   if (fd != -1)
+   {
+      vcsm_handle = dup(fd);
+
+      // FIXME: Sanity check which device that the fd actually relates to.
+      // For now we have to guess based on whether export is requested.
+      // (the main use case is from Chromium which will be requesting export).
+      if (want_export)
+         using_vc_sm_cma = 1;
+
+      goto out;
+   }
+
    if (want_export)
    {
-      if (fd == -1)
-         vcsm_handle = open( VCSM_CMA_DEVICE_NAME, O_RDWR, 0 );
-      else
-         // FIXME: Sanity check that the fd really is to vcsm-cma.
-         vcsm_handle = dup(fd);
+      vcsm_handle = open( VCSM_CMA_DEVICE_NAME, O_RDWR, 0 );
 
       if (vcsm_handle >= 0)
       {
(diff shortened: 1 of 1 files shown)


6by9:
Generated based on comments on https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=250730

Tested with raspistill on 64bit kernel, and MMAL no longer complains on startup, although raspistill doesn't use zero copy by default anyway.

pelwell:
Can you briefly explain the logic behind the device name use order?:
* export => vcsm before vcsm-cma
* !export => vcsm-cma before vcsm

6by9:
- Export is only supported by vcsm-cma.
- There are some caching operations that are currently only supported by vcsm.
- /dev/vcsm doesn't exist on 64bit systems.

- If the client has requested export, we have to go for vcsm-cma.
- If the client hasn't requested export then we currently default to vcsm so all the caching stuff works.
- On a 64bit system the open of vcsm will fail, therefore currently we end up with nothing. This patch changes it so that we try vcsm-cma and should get something.

Hopefully I'll get time to sort out the caching stuff in vcsm-cma, at which point we can hopefully deprecate vcsm totally, and this all disappears.

pelwell:
1. The handling of using_vc_sm_cma in the dup(fd) case looks strange - what's the relationship?
2. The dup(fd) case appears three times where one ought to be enough.

6by9:
1. Strange in what way? Passing in an fd is mainly a fudge to get around Chromium's sandboxing. Various sandboxes can't open devices, but they can accept file handles passed in from outside. Seeing as we can't identify which device the provided fd relates to, we have to rely on the same logic on which device we're talking to as when the original was created.

2. Yes it could be refactored into a dup case, and an opening case. I'll do that.

pelwell:
1. Strange in the sense that the fd passed in can be used with using_vc_sm_cma = 1 with export = 1 and export = 0, depending on what devices are available. Tidying up 2 will probably resolve that.

host_applications/linux/libs/sm/user-vcsm.c
```diff
@@ -273,13 +273,22 @@ int vcsm_init_ex( int want_export, int fd )
...
+   {
+      vcsm_handle = dup(fd);
+
+      // FIXME: Sanity check which device that the fd actually relates to.
+      // For now we have to guess based on whether export is requested.
+      // (the main use case is from Chromium which will be requesting export).
+      if (want_export)
+         using_vc_sm_cma = 1;
+
+      goto out;
```
pelwell: Is it OK to bypass the assignment to vcsm_page_size?

6by9: Nope. Fixed.

PR #585, opened 2019-10-22
Remove legacy pkgconfig to avoid Mesa conflicts

psyke83:
When building software targeting Pi, it's common practice to use
the pkgconfig to assist building by specifying PKG_CONFIG_PATH.

Unfortunately, the presence of the legacy pkgconfig that conflict with
Mesa's libraries will cause issues for users that build against the
firmware pkgconfig, but wish to target Mesa. (Pi 4B or earlier revisions
using the f/kms overlay).

Although a possible workaround may be to omit the custom PKG_CONFIG_PATH
of the firmware pkgconfig during build, there are plausible use cases in
which to target a software build against Mesa whilst also leveraging
the non GL-specific firmware libraries (e.g. MMAL for hardware codec support
that works on the FKMS overlay, or BCM_HOST for basic board detection).

Note: this conflict is already resolved via the Raspbian packages by deliberately
omitting the conflicting .pc files during the packaging stage; see https://github.com/raspberrypi/firmware/issues/857


popcornmix:
I think I'm okay with this. @6by9 @pelwell ?

6by9:
Packaging is not something I can claim knowledge about.
It looks OK to me, but does it leave legacy platforms with no way to include these modules?

popcornmix:
They should use brcmegl.pc brcmglesv2.pc brcmvg.pc for legacy packages.

psyke83:
Usage of the `brcm` prefix for the vendor libraries has been enforced since firmware version `1.20170811` - the first Raspbian stretch firmware release, which dropped the old library names that  interfere with Mesa. Most popular emulators and source ports that support the Pi have been updated to link against the `brcm` libraries - by necessity - for quite some time.

OT: For RetroPie, we're currently working on implementing support for 4B, whilst also adding optional support for users to target Mesa builds even on the older boards if desired. For all Mesa builds, we're aiming to run as much software as possible without X (via SDL2's KMSDRM driver, RetroArch's KMS driver, etc.). Normally we would avoid using `rpi-update` altogether, but there are some quite serious vsync-related tearing issues on both 3B and 4B, so we need to use `rpi-update` to troubleshoot these issues ahead of the staggered Raspbian package updates. Not having to worry about the legacy pkgconfig causing conflicts after `rpi-update` would be useful for anyone who relies on the firmware pkgconfig for building software against Mesa.

Issue #590, opened 2019-11-07
replace arch in buildme with "uname -m"


solskogen:
**Describe the bug**
Not every distribution comes with "arch", which is used in buildme. Replace it with something more portable, like "uname -m"

**To reproduce**
List the steps required to reproduce the issue.

**Expected behaviour**
Successfull build on Arch Linux and Manjaro

**Actual behaviour**
Since arch isn't installed, the build fails.

pelwell:
Submit a Pull Request - for a simple change like that you can do it all in the GitHub GUI.

solskogen:
Done!

PR #591, opened 2019-11-07
Replace 'arch' with uname

solskogen:


JamesH65:
Since `man arch` says "arch - print machine hardware name (same as uname -m)" this LGTM.

popcornmix:
@solskogen can you explain why?
Are you running this on a distribution that doesn't support ``arch``?

solskogen:
Not every distribution comes with "arch", which is used in buildme.

popcornmix:
Which distribution are you using that doesn't have arch?

solskogen:
Arch Linux ARM and Manjaro (which is based on Arch) doesn't come with 'arch', but when replaced with $(uname -m) it works perfectly.

pelwell:
See https://github.com/raspberrypi/userland/issues/590 for context.

pelwell:
LGTM.

PR #594, opened 2019-11-14
Update the GPS version to 23

JamesH65:
Buster now has gpsd version 23, (up from 22)

This PR updates the camera code to reflect the change.

### DIFF
--- a/host_applications/linux/apps/raspicam/gps.h
+++ b/host_applications/linux/apps/raspicam/gps.h
@@ -1,48 +1,7 @@
 /* gps.h -- interface of the libgps library */
 /*
  * This file is Copyright (c) 2010 by the GPSD project
- *
- * 			COPYRIGHTS
- *
- * Compilation copyright is held by the GPSD project.  All rights reserved.
- *
- * GPSD project copyrights are assigned to the project lead, currently
- * Eric S. Raymond. Other portions of the GPSD code are Copyright (c)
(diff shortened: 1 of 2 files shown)


JamesH65:
Note, the change to gps.h is simply a copy of the v23 version over the top of the v22 version.

6by9:
Do we want to just do this simplistic bump the version? It'll now be broken on Stretch.

I haven't looked at what symlinks or automagic mangling of the library name happens in dlopen, but it'd be nice to be able to load a generic library filename and then check against supported versions. Don't know if that is actually possible though.

JamesH65:
> Do we want to just do this simplistic bump the version? It'll now be broken on Stretch.
> 
> I haven't looked at what symlinks or automagic mangling of the library name happens in dlopen, but it'd be nice to be able to load a generic library filename and then check against supported versions. Don't know if that is actually possible though.

I did wonder about that, but on the whole we only have support for the current version of the distro, ie backward rather than forward compatibility. Might be worth asking @XECDesign though.

XECDesign:
Stretch is unlikely to get libraspberrypi0 updates. If you would like to provide an option for people to build it themselves, you could use ifdefs or dynamically load the appropriate .so at runtime and deal with whatever API differences might be relevant.

Personally, I'm in the "we don't support old releases" camp.

6by9:
Fine, ditch libgps.22

PR #598, opened 2019-11-27
Fix confusing typo swapping _MAP and _ALLOC

imbens:
VMCS_SM_IOCTL_HOST_WALK_PID_ALLOC and VMCS_SM_IOCTL_HOST_WALK_PID_MAP had been swapped, causing vcsm_status to yield unexpected results.


popcornmix:
Looks good. @6by9 okay?

6by9:
Not a function I've ever used, but looks reasonable.

PR #602, opened 2019-12-16
vmcs_host/vcilcs: build only on 32 bit

daggs1:
vcilcs requries libbrcmEGL, libbrcmGLESv2 and libkhrn_client which have
only 32 bit versions currently, so disable it when building aarch64

Signed-off-by: Dagg Stompler <daggs@gmx.com>


pelwell:
That looks plausible. Any comments, @6by9 ?

6by9:
Yes, looks fine.
I'm not totally clear what the real issue is - it all compiles fine for me, so I guess it must be creating a package or similar.

daggs1:
> Yes, looks fine.
> I'm not totally clear what the real issue is - it all compiles fine for me, so I guess it must be creating a package or similar.

I'm using buildroot to generate a pure 64 bit image, I think in your case, you setup is mixed x32 and x64

PR #622, opened 2020-05-08
raspistill: Adjust framerate limits on long exposures.

naushir:
Need to reduce the min fps limit to 0.005 for 200s exposures
available in HQ cam.

Not entirely sure why we have different limits based on exposure, but I have kept that as-is.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiStill.c
+++ b/host_applications/linux/apps/raspicam/RaspiStill.c
@@ -872,7 +872,7 @@ static MMAL_STATUS_T create_camera_component(RASPISTILL_STATE *state)
    if(state->camera_parameters.shutter_speed > 6000000)
    {
       MMAL_PARAMETER_FPS_RANGE_T fps_range = {{MMAL_PARAMETER_FPS_RANGE, sizeof(fps_range)},
-         { 50, 1000 }, {166, 1000}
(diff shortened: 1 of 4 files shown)


naushir:
@6by9, @JamesH65, this is why we were limited to 20s exposures for HQ cam.

6by9:
> Not entirely sure why we have different limits based on exposure, but I have kept that as-is.

It's the only way of switching into the longer exposure time mode on ov5647.

6by9:
LGTM

6by9:
Except it needs to be done in the other apps too
https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiStillYUV.c#L648 
https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiVid.c#L1604
https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiVidYUV.c#L924

naushir:
> Except it needs to be done in the other apps too

Done.

6by9:
I'm happy now.

JamesH65:
LGTM

JamesH65:
@popcornmix Another userland change to get back to the main build.

PR #623, opened 2020-05-09
raspicam: Adjust framerate limits on long exposures for stills port.

naushir:
Need to reduce the min fps limit to 0.005 for 200s exposures
available in HQ cam.

commit 27db31a8a828c095cb3e35f65d731c7628c587d7 fixed this for the video
port, but we need to do the same for the stills port as well.


naushir:
@6by9 and @JamesH65, did not realise that we had fps settings per port :(

6by9:
The change for the stills port in Raspivid and Raspividyuv won't do anything, but may as well make them all the same.
LGTM

PR #629, opened 2020-05-28
Fix a compiler warning

RaspberryPiFan:
This fixes the warning `format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘size_t’ {aka ‘long unsigned int’}`


JamesH65:
https://stackoverflow.com/questions/2524611/how-can-one-print-a-size-t-variable-portably-using-the-printf-family would seem to indicate %zu is appropriate for architecture agnostic code (ie 32 and 64)

Were you compiling on a 64bit system?

RaspberryPiFan:
Yes, I compiled using an ARM64 OS, I fixed the issue now.

popcornmix:
Looks okay - thanks.

PR #638, opened 2020-06-24
Fix cast warning on vcos_llthread_current

jc-kynesim:
Current header generates warnings of the form "warning: cast from
function call of type ‘pthread_t’ {aka ‘long unsigned int’} to
non-matching type ‘VCOS_LLTHREAD_T *’ {aka ‘struct VCOS_LLTHREAD_T *’}
[-Wbad-function-cast]".  Fix that.


6by9:
LGTM.

PR #645, opened 2020-07-27
Remove vcfiled/vcfilesys, and a couple of MMAL cleanups

6by9:
vcfiled/vcfilesys removal as per #631 #635 #644

A couple of minor MMAL cleanups whilst I was creating a userland PR. One for debug builds of the examples, and one so RaspiStillYUV matched Raspstill.


pelwell:
Looks good to me. I'll just give popcormix a chance to reprieve vcfiled.

PR #655, opened 2020-11-08
Make explicit that libfdt should be built static

crabbedhaloablution:
To avoid issues such as #333 , make explicit that libfdt should be built as static.
This will allow raspberrypi-userland to more easily integrate with distributions which automatically add  -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_STATIC_LIBS:BOOL=OFF to their CMake flags.

### DIFF
--- a/opensrc/helpers/libfdt/CMakeLists.txt
+++ b/opensrc/helpers/libfdt/CMakeLists.txt
@@ -1,6 +1,6 @@
 include_directories(.)
 
-add_library(fdt
+add_library(fdt STATIC
             fdt.c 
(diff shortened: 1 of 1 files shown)


HinTak:
I don't think it is advisable to guess what distributions need or do not need, unless you represent one of such distributions.

crabbedhaloablution:
I'm working on a respin of the Gentoo ebuild. The default behavior of Gentoo matches that seen in #333 .
The change is null for anyone else. I see no reason why it shouldn't be merged.

PR #659, opened 2020-11-16
Only install .pc files for which we build a .so.

crabbedhaloablution:
Many .so files are not built on arm64, but currently we install
the .pc files anyway. Fix by including the .pc files in the ARM64
conditional.


-
```diff
@@ -0,0 +1,25 @@
...
+ get_filename_component(VIDEOCORE_ROOT . ABSOLUTE)
+@@ -122,7 +124,7 @@ include_directories("${PROJECT_BINARY_DIR}")
+ include(FindPkgConfig QUIET)
+ if(PKG_CONFIG_FOUND)
+ 	# Produce a pkg-config file
+-	foreach(PCFILE bcm_host.pc brcmegl.pc brcmglesv2.pc brcmvg.pc vcsm.pc mmal.pc)
++	foreach(PCFILE IN LISTS PKGCONFFILES)
+ 		configure_file("pkgconfig/${PCFILE}.in" "${PCFILE}" @ONLY)
+ 		install(FILES       "${CMAKE_CURRENT_BINARY_DIR}/${PCFILE}"
+ 			DESTINATION "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig")
```
XECDesign, line 25: I'm guessing this patch was added unintentionally.

crabbedhaloablution: Yep, fixed.

PR #660, opened 2020-11-23
Fixup code for consistency

PokeCodec:
As well as formatting


6by9:
The standard indentation is 3 spaces per layer, so this makes it slightly more of a mixture.

The correction of return values, and changing len to unsigned are fine.

PokeCodec:
It is now 3 indentation spaces per layer.

PokeCodec:
@6by9 any updates?

PR #661, opened 2020-11-27
all host_applications: remove non-existent projects

twoerner:
The ALL_APPS symbol will optionally build an additional set of projects,
however, several of them don't exist anymore. Remove them from the list of
ALL_APPS.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>


pelwell:
`vcdbg` and `edidparser` do exist, but only in our internal firmware repo (of which this repo is a subset). `libs/elftoolchain` has gone though, so you can delete that CMakeLists line.

twoerner:
Ah, thanks for the explanation. That explains why several apparently removed  projects were still referenced in the CMakefile.txt. I'll prepare an update.

twoerner:
Updated.

pelwell:
That looks good to me.

@popcornmix ?

PR #663, opened 2020-11-30
Add man-pages for most commands

waveform80:
I've spent a little time going through the excellent raspberry pi documentation site in order to build some man-pages for most of the utilities in the suite. Note that not all utilities are covered here (notably vchiq_test, mmal_vc_diag are missing) but I think it covers the most commonly used utilities.

I've also included some auxilliary man-pages (in section 7) for information relevant to utilities like vcmailbox (covering things like the mailbox info from the wiki-page, and the OTP registers from the documentation site).

I'm happy to make any stylistic / content changes required. If you want to preview how they look currently, you can simply check out the branch and run "man" on the file in question.

### DIFF
--- a/host_applications/linux/apps/dtmerge/CMakeLists.txt
+++ b/host_applications/linux/apps/dtmerge/CMakeLists.txt
@@ -18,3 +18,4 @@ add_executable(dtmerge dtmerge.c)
 target_link_libraries(dtmerge dtovl)
 
 install(TARGETS dtmerge RUNTIME DESTINATION bin)
+install(FILES dtmerge.1 DESTINATION man/man1)
--- a/host_applications/linux/apps/dtmerge/dtmerge.1
+++ b/host_applications/linux/apps/dtmerge/dtmerge.1
@@ -1,0 +1,72 @@
+.TH DTMERGE 1
+.
+.SH NAME
+dtmerge \- merge an overlay or parameters into a base device-tree
+.
+.
+.SH SYNOPSIS
+.SY dtmerge
+.OP \-d
+.I base-dtb
+.I merged-dtb
+.I overlay-dtb
+.RI [ param=val \|.\|.\|.]
+.YS
+.
+.SY dtmerge
+.B \-h
+.YS
+.
+.
+.SH DESCRIPTION
+.B dtmerge
+is a command line utility for merging overlays and/or device-tree parameters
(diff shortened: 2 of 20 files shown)


JamesH65:
This LGTM, but not sure if merging it  here will work.  @popcornmix @XECDesign

waveform80:
Yes, this commit isn't enough to add the man-pages to the packaging of itself. In the Ubuntu packages I've been working on, these man-pages are currently just stuffed in the debian/ sub-dir with a libraspberrypi-bin.manpages file to install them, but I figured that's not as much use to you guys (or other packagers) as something a bit more integrated like this.

Still, if you think these pages would be better moved into their own sub-dir under the package root (e.g. "man/") with their own CMakeLists.txt to handle installing them, I'm happy to amend things?

XECDesign:
I think the issue is not so much the packaging (I can update that accordingly), but that this repo is derived from an internal repo.

waveform80:
Ah, I see - so this'll need comitting to the private then cherry-picked thru to the public (or something like that).

JamesH65:
@XECDesign If you are happy with this, are you the person to do the cherry pick? Or @popcornmix?

XECDesign:
That would be @popcornmix . I'll just need a nudge at some point to update the packaging side of things.

andrum99:
Hi folks. Just to say that these man pages are already present on Ubuntu 21.04, but are currently not in Raspberry Pi OS. If this could get integrated, then cherry picked to your internal repo from which the Raspberry Pi OS packages are built that would be good. It would be rather unfortunate if Raspberry Pi OS ended up without the man pages, given that Ubuntu has them.

For those that aren't aware, the author of this PR works on Raspberry Pi-specific stuff for Canonical.

andrum99:
> Hi folks. Just to say that these man pages are already present on Ubuntu 21.04, but are currently not in Raspberry Pi OS. If this could get integrated, then cherry picked to your internal repo from which the Raspberry Pi OS packages are built that would be good. It would be rather unfortunate if Raspberry Pi OS ended up without the man pages, given that Ubuntu has them.
> 
> For those that aren't aware, the author of this PR works on Raspberry Pi-specific stuff for Canonical.

Apologies - it occured to me that what I requested here may already have been done, since there is a new version of Debian, and therefore probably Raspberry Pi OS, in the works.

PR #665, opened 2020-11-30
Install libs into multi-arch friendly dirs

waveform80:
This patch tweaks the CMakeLists to install libraries into the default location instead of a hard-coded "lib". This is usually a multi-arch friendly location, e.g. `usr/lib/aarch64-linux-gnu`.

This may be too invasive a change for Raspberry Pi OS, where the location is already well established, but I figure the PR (even closed) might be useful for other distros looking to package it.


dwrobel:
Have you tried to look at https://github.com/raspberrypi/userland/pull/650?

PR #681, opened 2021-03-13
Update RaspiVid.c

TZubiri:
Changed a comment to clarify an implied condition.

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -1097,7 +1097,7 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
             if (sockListen >= 0)//regardless success or error
                close(sockListen);//do not listen on a given port anymore
          }
-         else//if (pState->netListen)
+         else//if (! pState->netListen)
          {
             if(0 <= (sfd = socket(AF_INET, socktype, 0)))
             {


HinTak:
I think you change to `//else from "if (pState->netListen)"` , as both the old and your suggested change is confusing.

pelwell:
Just delete the comment - it doesn't really add anything.

TZubiri:
@HinTak That's an acceptable alternative.
@pelwell , it's supposed to help with brace matching, it helped at least me to read the code

lurch:
IMHO the change in this PR is inappropriate, as all the other `else` blocks in this function are commented in a similar style, e.g.
```c
            if (sockListen >= 0)
            {
                ...
            }
            else//if (sockListen >= 0)
            {
                ...
            }
```

so if this was changed from
```c
        if (pState->netListen)
        {
            ...
        }
        else//if (pState->netListen)
        {
            ...
        }
```
to instead say `else//if (! pState->netListen)` that'd make it inconsistent with all the other code-comments.
Looking at the git history, this "unusual" commenting style was added in #359 

I agree with @pelwell and @HinTak that it might be better to just delete (all) of these potentially confusing comments.

TZubiri:
@lurch That's a good point, it would make it inconsistent and that would be confusing. Closing.

Regarding deleting these comments altogether, this is just one man's opinion, but it seems like you would be taking away the personality of the code in order to make it uniform. I find it's ok for there to be some level of inconsistency in a collaborative project, even if they impact the developer.

Consider ssh -p vs scp -P, yes it's slightly annoying, but that's how you can tell that they were made by different people, perhaps they are even different conventions, each with their own history and adherence. On the other hand if the inconsistency impacts the user, they should be sheltered from it, for example if there's close buttons on upper-right on some screens and upper-left on others.

The user should be shielded from complexity, not so the designer who should embrace it.

PR #683, opened 2021-03-18
Prefer BCM2711 nomenclature over BCM2838

lurch:
I think when this code was originally written we weren't sure whether we'd be referring to the chip as BCM2838 or BCM2711. But we seem to have firmly settled on BCM2711 by now :slightly_smiling_face:


popcornmix:
Seem okay to me.

JamesH65:
LGTM

PR #684, opened 2021-03-19
Update board revision codes

lurch:
...with info from https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md


JamesH65:
LGTM

6by9:
LGTM too.
I don't see any arrays that reference the type that would do an invalid lookup with this.

PR #689, opened 2021-04-25
Remove debug-related code from vc_image_structs.h

Alvarito050506:
The removed lines currently cause the compilation of programs that depend on `vc_image_structs.h` to fail, since `helpers/debug_utils/debug_writer.h` does not exist, nor do the `DEBUG_*` types.

### DIFF
--- a/interface/vctypes/vc_image_structs.h
+++ b/interface/vctypes/vc_image_structs.h
@@ -34,8 +34,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "interface/vcos/vcos_stdint.h"
 #include "interface/vcos/vcos_attr.h"
 
-#include "helpers/debug_utils/debug_writer.h"
-
 #include "interface/vctypes/vc_image_types.h"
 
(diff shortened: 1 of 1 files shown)


popcornmix:
@6by9 looks like one of your commits - do we use this?

6by9:
vc_image_structs.h was broken out so that it could be used from userland without needing the full vc_image stuff.

```vcdbg``` is the most likely candidate to be using it, but I thought your magic scripts only picked up files that are actually used by the build. I'm surprised we need almost any of these vc_image headers for the public userland repo.

6by9:
```rm -rf helpers/vc_image``` and ./buildme runs quite happily.
Likewise vcinclude/vc_image_types.h appears unused.

(helpers/v3d/v3d_common.h appears unused too)

Alvarito050506:
Hello,

This file is unused by the userland itself, but some programs depend on it. I came here from a [forum post](https://www.raspberrypi.org/forums/viewtopic.php?f=67&t=234404#p1434281) by @6by9, and while it's stated that the program shouldn't be used as a "good example", I thought removing the "bad" lines here would be useful.

PR #691, opened 2021-05-13
vcsm: properly handle too many CMA allocation attempts

### LINKED ISSUE #690: Segfault in vcsm-cma when allocating more than 512 buffers
Using the VCSM library (via the CMA path), when calling `vcsm_malloc_cache` more than 512 times, the last allocation crashes with a SEGFAULT.

This is most likely due to `vcsm_payload_list_get()` returning a NULL-pointer and `vcsm_malloc_cache` does not check for such a case and unconditionally [writes to it](https://github.com/raspberrypi/userland/blob/45a0022ac64b4d0788def3c5230c972430f6fc23/host_applications/linux/libs/sm/user-vcsm.c#L488).

doe300:
Now will properly report errors instead of crashing due to writing to NULL pointer, fixes #690.

### DIFF
--- a/host_applications/linux/libs/sm/user-vcsm.c
+++ b/host_applications/linux/libs/sm/user-vcsm.c
@@ -485,6 +485,13 @@ unsigned int vcsm_malloc_cache( unsigned int size, VCSM_CACHE_TYPE_T cache, cons
                       );
 
       payload = vcsm_payload_list_get();
+      if (!payload)
+      {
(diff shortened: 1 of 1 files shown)


6by9:
Yup, looks good to me. I'll wait in case anyone else has any input before merging.

pelwell:
My only objection is a stylistic one. I can see that there are multiple styles of whitespace use within this file, but at the very least any changes should be self-consistent. My preference is to not put additional whitespace inside the function calls, i.e. your `munmap(...)` call is good and your `vcsm_free( ... )` call is bad,

pelwell:
I'm happy with that.

PR #692, opened 2021-05-13
EGL: Move EGL_DISPMANX_WINDOW_T definition in a separate header

caramelli:
In order to build using a generic eglplatform.h (from EGL-Registry), this change simply moves the EGL_DISPMANX_WINDOW_T definition into a separate header while remaining compatible with the legacy eglplatform.h
This is similar to the native window definition used for the DRM EGL platform with the gbm.h header, or also similar to the native window definition used for the Wayland EGL platform with the wayland-egl.h header

Nicolas Caramelli

### DIFF
--- a/host_applications/linux/libs/bcm_host/include/bcm_host.h
+++ b/host_applications/linux/libs/bcm_host/include/bcm_host.h
@@ -48,6 +48,7 @@ unsigned bcm_host_get_peripheral_size(void);
 unsigned bcm_host_get_sdram_address(void);
 
 #include "interface/vmcs_host/vc_dispmanx.h"
+#include "interface/vmcs_host/vc_dispmanx_egl.h"
 #include "interface/vmcs_host/vc_tvservice.h"
 #include "interface/vmcs_host/vc_cec.h"
 #include "interface/vmcs_host/vc_cecservice.h"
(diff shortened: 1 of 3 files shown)


popcornmix:
`EGL_DISPMANX_WINDOW_T` (and firmware gl) is something we are moving away from, so we don't want to spend significant effort on it.
But this looks like a fairly painless change, so I'm not against it. @6by9 ?

@caramelli are you still using firmware gl in preference to arm side mesa gl? Any reason why?

caramelli:
I'm using both: the Mesa implementation is working well, but I find this legacy implementation interesting and I'm still playing with the samples based on Dispmanx or with softwares that still support it.
I'm using this proposed change for building EGL/OpenGL softwares on my system because I think it makes sense to be able to use a generic and unique EGL header file for both implementations (even if EGL_DISPMANX_WINDOW_T is now deprecated)

caramelli:
Do you still have reservations about this change?

popcornmix:
@6by9 any objection?

caramelli:
We generally avoid defining the EGL native window type directly in eglplatform.h
I really consider the Raspberry Pi as a reference board and this is a change to be consistent with what is usually done.
But if I'm wrong, feel free to comment.

6by9:
Sorry, I'd missed the ping first time round.
As long as all the hello_pi examples still build, then I'm fine with it.

It may trip up gstreamer as well, but that's probably the only app that is likely to be using it that we sort of care about.

caramelli:
Thanks for your feedback.
I have successfully built and run the hello_pi examples, the GStreamer gstgl library based on Dispmanx EGL, and others like SDL2.
But if you encounter any issues, do not hesitate to report it.

Issue #695, opened 2021-06-23
CM4 host


gregoryjaooi:
Problem : 
Some users are using CM4 and the source codes are using bcm_host_is_model_pi4() function call to
check for BCM2711 to differentiate from other older Pi Models.
As such, the same application has error/issues running on CM4 (0x14) or PI400 (0x13) while
on PI4 (x11) there are no issues at all.

/* Returns the type of the Pi being used
*/
int bcm_host_is_model_pi4(void)
{
   return bcm_host_get_model_type() == 0x11 ? 1 : 0;
}

Or may be there should be a better work around that I am not aware of.

pelwell:
@popcornmix may have some thoughts on the matter, but as you say I think this function makes more sense as a way of detecting any member of the Pi 4 family, i.e. 4B, 400 and CM4. This is equivalent to checking if the processor is a BCM2711, so I think it should be implemented in that way. While looking at this I also noticed that the list of processor types include BCM2838 - our guess at the Broadcom name for what turned out to be called BCM2711. `BCM_HOST_PROCESSOR_BCM2711` should be included as an alias.

pelwell:
See https://github.com/raspberrypi/userland/pull/696.

gregoryjaooi:
This issue can be closed as CM4
is running very well now after raspi-update.
Thanks.

PR #696, opened 2021-06-23
bcm_host: Recognise all Pi 4 variants, add BCM2711

pelwell:
bcm_host_is_model_pi4 should be equivalent to a test that the processor
type is BCM2711, so implement it as such (having added BCM2711 as a
recognised processor type and deprecating the name BCM2838).

See: https://github.com/raspberrypi/userland/issues/695

Signed-off-by: Phil Elwell <phil@raspberrypi.com>

### DIFF
--- a/host_applications/linux/libs/bcm_host/bcm_host.c
+++ b/host_applications/linux/libs/bcm_host/bcm_host.c
@@ -265,11 +265,11 @@ int bcm_host_get_model_type(void)
    return model_type;
 }
 
-/* Returns the type of the Pi being used
+/* Test if the host is a member of the Pi 4 family (4B, 400 and CM4)
 */
 int bcm_host_is_model_pi4(void)
 {
-   return bcm_host_get_model_type() == 0x11 ? 1 : 0;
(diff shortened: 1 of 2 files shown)


6by9:
LGTM unless we want to make the distinction between pi4 and bcm2711. I'm not sure of my view on that one seeing as the user can just as easily check the model number directly without a function.

@jc-kynesim was asking about this the other day as it meant one of his checks in either VLC or Chromium failed on Pi400.

pelwell:
I don't really understand why this function exists, but the few uses I can find are essentially checks of the SoC capabilities, hence this change. I certainly don't think we should add any more "helpers".

jc-kynesim:
I think I was the one that asked for this, as you say, for a capability check so I could look for the right decode h/w with the expectation that "pi4" would cover all variants that were essentially the same and I didn't have to update my code as the system changed.

pelwell:
Ha - I've just found an internal Merge Request from @popcornmix that appeared while I was on holiday doing exactly the same thing. I think that counts as consent.

gregoryjaooi:
Thanks guys! I am new to this ... but glad that the issue of detecting pi4 family is being addressed.

PR #700, opened 2021-07-09
Add bash programmable completion for vcgencmd

rgm3:
Package the `vcgencmd-completion.bash` file as `/usr/share/bash-completion/completions/vcgencmd` to install.
The integration tests don't need to be shipped, but may help maintenance when updating the completion for future changes.


rgm3:
The CMakeLists.txt file may want:
```
install(FILES vcgencmd-completion.bash DESTINATION bash-completion/completions/vcgencmd)
```

But I'm not sure.  I'd be happy to update the PR if that seems correct.

popcornmix:
This seems useful.
@XECDesign how would this be best used in RPiOS? Would you install the file directly in `/usr/share/bash-completion/`?
Or just leave it to the user to install?

XECDesign:
I'd modify the packaging to install it appropriately.

PR #713, opened 2021-11-23
raspicam: Check system is running legacy camera stack

davidplowman:
If the system is obviously configured for libcamera (for example) by
dint of /dev/video0 being given over to "unicam", complain and abort
immediately. In all other circumstances maintain previous behaviour.

(There seemed to be no obvious place for common code so it is
duplicated. There's not much.)

Signed-off-by: David Plowman <david.plowman@raspberrypi.com>

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiHelpers.c
+++ b/host_applications/linux/apps/raspicam/RaspiHelpers.c
@@ -33,6 +33,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <unistd.h>
 #include <stdint.h>
 
+#include <fcntl.h>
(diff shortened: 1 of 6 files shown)


pelwell:
I like the detection method, but put `check_camera_stack` in `RaspiHelpers.c`.

6by9:
A USB camera connected to the system could enumerate as /dev/video0 before bcm2835-unicam loads, but so be it.
If it is configured for the libcamera stack, then the firmware won't detect any cameras anyway, so creating the camera component will fail.

davidplowman:
Ah thanks, didn't spot RaspiHelpers.c! Second attempt incoming...

pelwell:
That works for me.

6by9:
LGTM

PR #714, opened 2021-11-24
mmal: Hack to make mmal core register VideoCore components

6by9:
The default linker option with Bullseye appears to now set
as-needed, so as the core doesn't call into mmal_vc_client
it is viewed as unnecesary, missing that mmal_vc_client has
a constructor that registers functions with the core.

Move the registration of the component supplier to the core_init.
The linker is therefore satisfied that mmal_vc_client is used,
and actually links to it.

This allows Picamera to work again.

https://github.com/waveform80/picamera/issues/697

Signed-off-by: Dave Stevenson <dave.stevenson@raspberrypi.com>

### DIFF
--- a/interface/mmal/core/CMakeLists.txt
+++ b/interface/mmal/core/CMakeLists.txt
@@ -11,7 +11,7 @@ add_library (mmal_core ${LIBRARY_TYPE}
    mmal_clock.c
 )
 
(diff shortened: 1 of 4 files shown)


popcornmix:
Okay by me.

PR #717, opened 2021-12-04
Install manual pages as per GNU Coding Standards

stefwalter:
Install manual pages as per GNU Coding Standards
    
Includes GNUInstallDirs and uses MANDIR (instead of hardcoded man)
to install manual pages.
    
See: https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

This patch depends on a similar #650 to be merged first.


stefwalter:
@dwrobel This is a similar patch to #650, and you might choose to bring it into your branch for merging at the same time.

PR #718, opened 2021-12-12
raspivid:Turn off buffering on fwrite(3)

dsugisawa-mixi:
This change makes low-latency udp streaming more stable in the BCM2835:Zero W environment.
(BCM2835 provides a 128KB system L2 cache, which is used primarily by the GPU)
 
+ script
```
./build/bin/raspivid -w 640 -h 360 -o udp://192.168.15.202:50002 -fps 30 --codec H264 -n -t 0
```

### DIFF
--- a/host_applications/linux/apps/raspicam/RaspiVid.c
+++ b/host_applications/linux/apps/raspicam/RaspiVid.c
@@ -1117,6 +1117,10 @@ static FILE *open_filename(RASPIVID_STATE *pState, char *filename)
 
          if (sfd >= 0)
             new_handle = fdopen(sfd, "w");
+            if (bcm_host_get_processor_id() == BCM_HOST_PROCESSOR_BCM2835)
+               // Turn off buffering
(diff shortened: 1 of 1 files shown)


pelwell:
It needs some explanation by the submitter beyond two statements and a patch, none of which are obviously related.

6by9:
https://www.cplusplus.com/reference/cstdio/setvbuf/
> _IONBF  No buffering: No buffer is used. Each I/O operation is written as soon as possible. In this case, the buffer and size parameters are ignored.

Nothing to do with CPU caches.

You're not saving much anyway as there is already an option for flushing after every packet with ```-fl```.
UDP supports a max 65507 bytes when over IPv4 (65535 - 8 byte UDP header - 20 byte IP header), and then then needs to be fragmented down to a typically max 1500 byte ethernet frame.

dsugisawa-mixi:
I agree,  and thanks your comments.

PR #723, opened 2022-03-23
raspicam: Fix raspicommonsettings_parse_cmdline

Krawlow:
Fix raspicommonsettings_parse_cmdline's definition to match declaration


pelwell:
Given that the Titanic (i.e. raspicam) is sinking (deprecated), moving this deck chair (correcting an error that is purely cosmetic) is a bit pointless.

Krawlow:
Nice comment ;) 

However, when inserting this code in a C++ environment, calling this function becomes an undefined reference. Moreover, the similar function "display_valid_parameters" from RaspiHelpers.h has a matching definition and declaration, thus, giving no problems when using it in C++.

pelwell:
Fair enough - as a kernel type it's easy to forget about C++.

PR #734, opened 2023-09-12
raspivid: Also flush PTS file if flush is enabled

skadisch:


6by9:
I would normally have wanted the conditional code on a separate line from the if, but I see that the imv path in the same function didn't get that, so I won't complain too much. The whole legacy camera stack is considered deprecated anyway.

LGTM.

PR #737, opened 2023-10-18
README: Update to make it clear that most code in this repo is deprecated

popcornmix:


popcornmix:
@6by9 @pelwell 

Just making it clearer that using code from here is probably a bad choice.

pelwell:
There is still at least one thing still built and installed from userland - the dtoverlay/dtparam commands. It/they should be moved into the utils repo - but your statement is arguably correct.

Otherwise, looks good to me.

macmpi:
Thanks for clarification.
It may need a bit more details though, as some [utils](https://github.com/raspberrypi/utils) ([`rapinfo`](https://github.com/raspberrypi/utils/blob/master/raspinfo/raspinfo) for instance) use `tvservice`, `vcgencmd` at least (may miss some others): can we instantiate which apps & libs are still supported?

popcornmix:
I suspect:
> If you are considering using code from here that relates to displays, video codecs, cameras or 3d then this is probably the wrong solution.

does distinguish the deprecated from supported apps.
tvservice is not supported (relates to displays).
vcgencmd is supported (has many commands, only the display related ones are gone).

raspinfo doesn't require tvservice (it's in a function called `display_info_legacy` which is only called when using the deprecated/legacy display driver).

macmpi:
> vcgencmd is supported

Much clearer when spelled-out (as for dtoverlay/dtparam). Thanks

README.md
```diff
@@ -1,3 +1,19 @@
+======================================================================
+
+# Update
+
+This repo is ancient and largely deprecated.
+
+It largely contains code using proprietary APIs to interface to the VideoCore firmware. We have since moved to standard linux APIs.
+
+Code from here is not installed on latest RPiOS Bookworm images.
```
XECDesign, line 9: Heads up that we do still ship quite a lot of stuff from this repo.

popcornmix: What is still shipped? vcgencmd and dtoverlay? Anything else?
Might be best to move useful stuff into raspberrypi/utils.

XECDesign: [full list here](https://github.com/raspberrypi/userland/files/13043417/files.txt)

XECDesign:
Okay, I'll push a final raspberrypi-userland update to apt as is for now and then address the rest after utils has been updated.

pelwell:
Yes, I was tacitly agreeing to the plan.

pelwell:
See https://github.com/raspberrypi/utils/pull/40 and https://github.com/raspberrypi/utils/pull/41.

popcornmix:
Updated the readme message to be more final about the closure of this repo.

nettnikl:
Hi, just to confirm: this is the reason for the conflict on Arch, Manjaro, etc... for the conflict of
**raspberrypi-utils** and **raspberrypi-userland-aarch64-git**, correct?
So raspberrypi-utils contains now everything from the other package, and we can safely remove raspberrypi-userland-aarch64-git (see https://github.com/raspberrypi/utils/pull/40).

PR #744, opened 2025-07-06
Fix conflicting types build error

jonaski:


6by9:
Please note the README for this repo

> This repo is ancient and deprecated.
> 
> It largely contains code using proprietary APIs to interface to the VideoCore firmware. We have since move to standard linux APIs.
> 
> V4L2, DRM/KMS and Mesa are the APIs you should be using.
> 
> The few useful tools from here (dtoverlay, dtmerge, vcmailbox, vcgencmd) have been moved to the raspberrypi/utils repo.
> 
> Code from here is no longer installed on latest RPiOS Bookworm images.
> 
> If you are using code from here you should rethink your solution.
> 
> Consider this repo closed.

If you have a very good reason to need to fix these up then please state it, otherwise this PR will be closed.

jonaski:
openSUSE and other distros supporting Raspberry Pi relies on it since the new utils vcgencmd does not work at all without the vcio driver which is not in upstream linux.
Do you have any idea why the vcio driver is not upstreamed in the linux kernel? Has anyone attempted to add it?
Where is the official repository for the driver? Is it this? https://maazl.de/project/vcio2/doc/

tpetazzoni:
I confirm that we have to carry the same patch in Buildroot to still allow building raspberrypi userland with GCC 15.x.

